PR #41, opened 2019-08-17
Petition to expunge continuation 

seanbw89:
Created radio component for the forms court type selection.
Added class name to select component to make the style in line with the rest of the form.
Removed checkbox and added radio buttons to form. 
Removed Fire-base timestampsInSnapshots as this is know the default

### DIFF
--- a/src/app/forms/inputs/radio.component.js
+++ b/src/app/forms/inputs/radio.component.js
@@ -1,0 +1,39 @@
+import React from "react";
+import { useCss } from "kremling";
+import { get } from "lodash";
+import { DataContainerContext } from "../data-container.component";
+
+export default function Radio(props) {
+  const [value, setValue] = React.useState(get(props.data, props.dataKey));
+  const scope = useCss(css);
+  const dataContext = React.useContext(DataContainerContext);
+
+  return (
+    <div className="web-form-input" {...scope}>
+      <div className="radio">{props.label}</div>
+      {props.options.map(option => (
+        <label key={option.value}>
+          {option.label}
+          <input
+            type="radio"
+            name={option.label}
+            value={option.value}
+            checked={value == option.value}
+            onChange={evt => setValue(evt.target.value)}
+            onBlur={handleBlur}
+          />
+        </label>
+      ))}
+    </div>
+  );
+
+  function handleBlur() {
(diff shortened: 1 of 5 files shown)


src/app/forms/petition-for-conviction/petition-for-conviction.pdf.component.js
```diff
@@ -121,14 +121,14 @@ export default class PetitionForConviction_Pdf extends React.Component {
             debugKey="caseCourtType"
             left="37.5%"
             top="36.5%"
-            dataKey="case.districtCourt"
```
joeldenning: We should do `dataKey` when we are showing data and `debugKey` when we have a container that calls `renderData()` inside of it. In this case, we should use `dataKey`.

src/app/forms/petition-for-conviction/petition-for-conviction.pdf.component.js
```diff
@@ -121,14 +121,14 @@ export default class PetitionForConviction_Pdf extends React.Component {
...
+            shouldShow={data.case.courtType === "District"}
           />
 
           <PositionedCheckmark
             debugKey="caseCourtType"
-            left="48%"
-            top="36.5%"
-            dataKey="case.justiceCourt"
```
joeldenning: same here about using `dataKey` instead of `debugKey`

src/firebase/firebase.js
```diff
@@ -18,7 +18,5 @@ if (!firebase.apps.length) {
 const auth = firebase.auth();
 
 const database = firebase.firestore();
-const settings = { timestampsInSnapshots: true };
-database.settings(settings);
```
joeldenning: Removing this fixes the console warning? Nice

src/app/forms/inputs/radio.component.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import { useCss } from "kremling";
+import { get } from "lodash";
+import { DataContainerContext } from "../data-container.component";
+
+export default function Radio(props) {
```
joeldenning: 👍

seanbw89:
I have made those changes you requested. And verified all the other keys were the correct ones as well.

PR #44, opened 2019-08-19
started on draft-court-order-conviction and created common form options

seanbw89:


src/app/forms/draft-court-order-conviction/draft-court-order-conviction.web.component.js
```diff
@@ -0,0 +1,70 @@
...
+        <TextInput dataKey="person.addressZip" label={__("zip")} />
+
+        <TextInput dataKey="person.dayPhone" label={__("phone")} />
+
+        <TextInput dataKey="person.email" label={__("email address")} />
+
+        <Radio
+          dataKey="case.courtType"
+          label="Court Type"
+          options={formCommonOptions.courtTypeOptions}
```
lukeschunk: When you say `formCommonOptions` here, they are common to a single case, right? maybe we should save this as `case.xxx`? What do you think?

src/app/forms/form-common-options/form-common-options.js
```diff
@@ -0,0 +1,192 @@
...
+        break;
+      case "Daggett":
+      case "Duchesne":
+      case "Uintah":
+        this.JudicialDistrict = "Eighth";
+        break;
+    }
+    return this.JudicialDistrict;
+  }
+};
```
lukeschunk: This is fine, but also another way to do this would be a map/object, ie: 
```
const judicialDistricts = {
  Daggett: 'Eighth',
  Duchesne: 'Eighth'
}
// then: 
judicialDistricts[whatever]
```

src/app/forms/form-common-options/form-common-options.js
```diff
@@ -0,0 +1,192 @@
...
+//             JudicialDistrict = "Seventh";
+//             break;
+//         case "Daggett":
+//         case "Duchesne":
+//         case "Uintah":
+//             JudicialDistrict = "Eighth";
+//             break;
+//     }
+//     return JudicialDistrict;
+// }
```
lukeschunk: do you want to leave this commented code in this PR?

src/app/forms/draft-court-order-conviction/draft-court-order-conviction.web.component.js
seanbw89: Form common options would be the options that go into the select, radio, or checkbox components as props. The counties, court type etc...

src/app/forms/form-common-options/form-common-options.js
seanbw89: Yes the reason I left that commented code in the PR is I wasn't sure what the best practice for something like this would be.

src/app/forms/draft-court-order-conviction/draft-court-order-conviction.web.component.js
lukeschunk: 👍 got it

src/app/forms/form-common-options/form-common-options.js
lukeschunk: gotcha. I think either can be good. personally, I like the individual exports better. 

then when you import you can do something like: 
```
 import { courtTypeOptions, countyOptions } from '/wherever'
```
There's one less level to dig into. Doing it this way can also be better for bundles splitting

src/app/forms/form-common-options/form-common-options.js
```diff
@@ -0,0 +1,192 @@
...
+  courtTypeOptions: [
+    { label: "District Court", value: "District" },
+    { label: "Justice Court", value: "Justice" }
+  ],
+  getJudicialDistrictFromCounty: function(value) {
+    switch (value) {
+      case "Box Elder":
+      case "Rich":
+      case "Cache":
+        this.JudicialDistrict = "First";
```
joeldenning: Instead of setting `this.JudicialDistrict`, just return `First`. That way you don't have a temporary variable exposed on formCommonOptions that shouldn't really be used by anyone importing formCommonOptions.

src/app/forms/form-common-options/form-common-options.js
```diff
@@ -0,0 +1,192 @@
+export const formCommonOptions = {
```
joeldenning: 👏 👏 

Nice work getting this into a shared state. Let's change it so that we export each thing individually instead of all in one big `formCommonOptions` object. For example:

```js
export const petitionerRepresentativeOptions = [...]

export const countyOptions = [...]

export const courtTypeOptions = [...]
```

src/app/forms/form-common-options/form-common-options.js
joeldenning: 👍 yeah that way might be a bit easier. However, if you want to throw an error when they are looking up a county that doesn't exist, the function with a switch/case/default would be able to do that whereas the object couldn't.

I think that either way works.

src/app/forms/form-common-options/form-common-options.js
joeldenning: Yeah individual exports are more common - let's go with those.

seanbw89:
Updated the form common options, and forms that were using them.

PR #40, opened 2019-08-15
Luke/drug form continued

lukeschunk:
Addresses: https://github.com/JustUtahCoders/utahexpungements.org/issues/11 

* TODO: positioning polish  

Screen Shot 2019-08-15 at 2 27 59 PM

PR #45, opened 2019-08-20
Continue court order conviction

seanbw89:
Changed radio from handleBlur to usEffect, 
Finished out the form with inputs and positioned strings


src/app/forms/draft-court-order-conviction/draft-court-order-conviction.pdf.component.js
```diff
@@ -0,0 +1,197 @@
...
+          </PositionedString>
+        </RenderPage>
+      </>
+    );
+  }
+}
+
+const farLeft = `11.28%`;
+const petRepLeft = `20.82%`;
+const barNumberLeft = `62.2%`;
```
lukeschunk: i really like these, I'd probably declare them at the top just after the imports though. that's typically the convention I've seen

src/app/forms/form-common-options/form-common-options.js
```diff
@@ -0,0 +1,94 @@
...
+    case "San Juan":
+      return "Seventh";
+      break;
+    case "Daggett":
+    case "Duchesne":
+    case "Uintah":
+      return "Eighth";
+      break;
+  }
+}
```
lukeschunk: This looks good 👍

lukeschunk:
also a couple merge conflicts (?)

seanbw89:
Fixed the merge conflicts

src/app/forms/inputs/radio.component.js
```diff
@@ -8,6 +8,10 @@ export default function Radio(props) {
   const scope = useCss(css);
   const dataContext = React.useContext(DataContainerContext);
 
+  useEffect(() => {
+    dataContext.setData(props.dataKey, value);
```
joeldenning: I think that calling `dataContext.setData` triggers a save to firebase, which is why we were only doing it on blur previously. Calling firebase after every letter they type might be too much. When the user is not logged in, `setData` just calls setState and is totally fine, though.

Was there a bug associated with calling it on blur? If not, let's change it back to use blur.

seanbw89: I don't know if this is a bug, but it definitely is not user friendly. Trying so show a input box based off this value. The input box won't show up until you focus on the next input box.

lukeschunk: can you queue into the answer from the previous question to conditionally show the next field? like: https://github.com/JustUtahCoders/utahexpungements.org/pull/40/files#diff-31e7e0f2491482b249912bec08d7368fR128

lukeschunk: ah - i see that wouldn't work in this case because it is a text input, not a select

joeldenning: With radio inputs, the value doesn't change nearly as often as text inputs. So I think it's fine to save to db every time the radio changes.

No need to change this.

lukeschunk: ignore all of my comments i was confused.

PR #53, opened 2019-09-10
Add "Court Order (Drug Conviction)" Form, fix some form control bugs

jenseng:
The new form is pretty straightforward (hopefully I didn't miss anything 🤞)

I came across the input/select/radio bugs while working on the new form, and discovered it affected existing forms as well. If you'd like I can split that into a separate pull request.


frontend/app/forms/inputs/checkbox.component.js
```diff
@@ -4,29 +4,24 @@ import { get } from "lodash";
 import { DataContainerContext } from "../data-container.component";
 
 export default function Checkbox(props) {
-  const [value, setValue] = React.useState(
-    get(props.data, props.dataKey, false)
```
jenseng: previously it never had the right initial value, since it was looking in props instead of context

frontend/app/forms/inputs/radio.component.js
```diff
@@ -1,16 +1,12 @@
...
 import { get } from "lodash";
 import { DataContainerContext } from "../data-container.component";
 
 export default function Radio(props) {
-  const [value, setValue] = React.useState(get(props.data, props.dataKey));
-  const scope = useCss(css);
   const dataContext = React.useContext(DataContainerContext);
-
-  useEffect(() => {
-    dataContext.setData(props.dataKey, value);
```
jenseng: same story here

frontend/app/forms/inputs/select.component.js
```diff
@@ -5,15 +5,16 @@ import { DataContainerContext } from "../data-container.component";
 
 export default function Select(props) {
   const dataContext = React.useContext(DataContainerContext);
-  const [value, setValue] = React.useState(
-    get(dataContext.data, props.dataKey)
```
jenseng: although it was looking in context, if the first render happened before the data loaded, you'd get undefined in your state

frontend/app/forms/inputs/checkbox.component.js
joeldenning: 👍 good catch -- we recently switched from props to context and must have missed this.

PR #54, opened 2019-09-11
Update on the group select for court addresses

seanbw89:
simplified the flow how a user would select a court based on previous code review.


frontend/app/forms/form-common-options/form-common-options.js
```diff
@@ -46,7 +48,38 @@ export const courtTypeOptions = [
   { label: "Justice Court", value: "Justice" }
 ];
 
-export function getJudicialDistrictFromCounty(value) {
+export function getJudicialDistrict(courtAddress, courtType) {
```
joeldenning: This function could be simplified significantly:

```js
export function getJudicialDistrict(courtAddress, courtType) {
  const courtList = courtType === "District" ? DistrictCourtList : JusticeCourtList;
  const court = courtList.find(court => court.options.some(option => option.value === courtAddress))
  const county = court.name.replace(" County", "")
  return {
    judicialDistrict: getJudicialDistrictFromCounty(county),
    county,
  }
}
```

frontend/app/forms/form-common-options/form-common-options.js
```diff
@@ -46,7 +48,38 @@ export const courtTypeOptions = [
...
+  var judicialDistrict, county;
+  if (courtAddress) {
```
joeldenning: When is `courtAddress` not being passed in? This function is returning `undefined` instead of an object when that happens, which seems a bit weird. Should we just ensure that courtAddress is always passed in, so we don't have to have this if statement?

frontend/app/forms/inputs/group-select.component.js
```diff
@@ -0,0 +1,37 @@
...
+  const scope = useCss(css);
+
+  return (
+    <div className="web-form-input" {...scope}>
+      <div className="select-label">{props.label}</div>
+      <select
+        value={value}
+        onChange={evt => dataContext.setData(props.dataKey, evt.target.value)}
+      >
+        <option value="nothingSelected">Please select an option.</option>
```
joeldenning: Do the PDFs still render when `nothingSelected` is the option that was chosen by the user?

frontend/app/forms/acceptance-of-service/acceptance-of-service.pdf.component.js
```diff
@@ -57,19 +57,17 @@ export default function AcceptanceOfService_Pdf({ renderData, data }) {
...
         />
 
-        <PositionedString
-          dataKey="case.county"
-          left="51.4%"
-          top={countyDistrictTop}
-        />
+        <PositionedString left="50.7%" top="36.65%" debugKey="county">
+          {data.case.courtAddress
+            ? getJudicialDistrict(data.case.courtAddress, data.case.courtType)
```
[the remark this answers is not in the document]
seanbw89: I have seen this, have been using it to return jsx. Think I may have just skipped over this because of the lack of jsx.

PR #55, opened 2019-09-12
Started the motion to waive fees form

seanbw89:


frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -3,21 +3,233 @@ import { Scoped } from "kremling";
...
+  { value: "defendantRespondent", label: "Defendant / Respondent" },
+  {
+    value: "defendantRespondentAttorney",
+    label: "Defendant / Respondent's Attorney"
   }
+];
+
+export default function MotionToWaiveFees_Web({ data }) {
+  return (
+    <Scoped css={css}>
```
joeldenning: In function components, it's preferred to use the `useCss()` hook instead of rendering the `<Scoped>` component. See https://kremling.js.org/api/use-css.html. Reason why is that it's easier to understand which dom elements have the css scope when you use hooks. See https://kremling.js.org/api/scoped-component.html#gotchas for how the `<Scoped>` component can be a bit confusing sometimes.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -3,21 +3,233 @@ import { Scoped } from "kremling";
...
+
+          <Select
+            dataKey="person.relation"
+            label="Relation"
+            options={waiveFeeRelationship}
+          />
+
+          {data.person.relation &&
+            data.person.relation.includes("Attorney") && (
+              <TextInput label="Bar Number" dataKey="person.attorney" />
```
joeldenning: Is this the first time that we are using `person.attorney` as a dataKey? If so, I'd prefer changing it to `person.attorney.barNumber`, so that if we need to capture the attorney's name or other info we have a whole object for the attorney instead of just a string barNumber.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -3,21 +3,233 @@ import { Scoped } from "kremling";
...
+          <small className="web-form-input">(Choose all that apply)</small>
+          <CheckBox dataKey="case.filingFee" label="Filing fee" />
+          <small className="web-form-input">(Refer to Civil Cover Sheet)</small>
+          {data.case.filingFee && (
+            <TextInput
+              label="Filing Fee Amount"
+              dataKey="case.filingFeeAmount"
+            />
+          )}
+          <CheckBox dataKey="case.ocapFee" label="OCAP fee" />
```
joeldenning: These checkboxes reminded me that we should really have the CheckBox component require that you pass in a `name` prop. The name prop groups checkboxes for the same question together.

Can you add that to the CheckBox component. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-name

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -3,21 +3,233 @@ import { Scoped } from "kremling";
...
+          )}
+        </Section>
+        <Section name="5. Dependents">
+          <p className="web-form-input">
+            The following people depend on me for support
+          </p>
+          <small className="web-form-input">
+            (Count Spouse, children or other dependents in your house hold.)
+          </small>
+          <TextInput label="Number of adults" />
```
joeldenning: These are missing dataKeys. Is that on purpose? Are you planning on adding the dataKeys later?

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
seanbw89: Thanks I will look into that. Was switching over from the class components because it looks like most of the other forms are moving in that direction.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
seanbw89: That makes sense. Will be more expressive when it comes to the dataKeys.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
seanbw89: Will do.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
seanbw89: Yes I was trying to get a rough layout of the form. Will be adding dataKeys as needed.

joeldenning: 👍

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.pdf.component.js
```diff
@@ -1,19 +1,611 @@
...
+      <RenderPage url="/static/forms/motion-to-waive-fees/Motion_to_Waive_Fees_and_Statement_Supporting-10.png">
+        <PositionedString debugKey="todaysDate" left="11.5%" top="30.6%">
+          {moment().format("L")}
+        </PositionedString>
+        <PositionedString debugKey="printedName" left="54.4%" top="33.6%">
+          {`${renderData("person.firstName")} ${renderData(
+            "person.middleName"
+          )} ${renderData("person.lastName")}`}
+        </PositionedString>
+      </RenderPage>
```
lukeschunk: this form is such a beast. normally i'd probably recommend we break this up into multiple components (maybe one per page) but In the interest of completing things I think this is 👍

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.pdf.component.js
```diff
@@ -1,19 +1,611 @@
...
+          left="40.7%"
+          top="31.68%"
+          shouldShow={data.person.relation === "defendantRespondent"}
+        />
+        <PositionedCheckmark
+          dataKey="person.relation"
+          left="40.7%"
+          top="33.08%"
+          shouldShow={data.person.relation === "defendantRespondentAttorney"}
+        />
```
lukeschunk: since all of these are based off of `person.relation` i might try to dry it up a bit. that said; since the positioning is different in all of them it might not be worth it.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -1,23 +1,533 @@
...
+        <TextInput dataKey="person.firstName" label="First name" />
+        <TextInput dataKey="person.middleName" label="Middle name" />
+        <TextInput dataKey="person.lastName" label="Last name" />
+        <TextInput dataKey="person.addressStreet" label="Street address" />
+        <TextInput dataKey="person.addressCity" label="City" />
+        <TextInput dataKey="person.addressState" label="State" />
+        <TextInput dataKey="person.addressZip" label="Zip" />
+        <TextInput dataKey="person.phone" label="Phone" />
+        <TextInput dataKey="person.email" label="Email" />
+        <small className="web-form-input">
```
lukeschunk: TIL `<small>`

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -1,23 +1,533 @@
...
+            <TextInput
+              label="Child Support"
+              dataKey="person.income.childSupport"
+            />
+            <TextInput
+              label="Payments from civil litigation "
+              dataKey="person.income.cilvilPayments"
+            />
+            <TextInput
+              label="Victim resitutuion"
```
lukeschunk: i believe there is a typo here on `restitution`

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -1,23 +1,533 @@
...
+        <TextInput
+          label="Other transportation costs"
+          dataKey="person.income.transportation"
+        />
+        <small className="web-form-input">
+          (public transportation, parking, etc.)
+        </small>
+        <TextInput label="Utilities" dataKey="person.income.utilities" />
+        <small className="web-form-input">
+          (such as electricitym gas, water, sewage, garbage)
```
lukeschunk: typo here electricitym

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
seanbw89: 👍

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.pdf.component.js
seanbw89: Yeah this is a monster form. I think I may break this into smaller components.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
seanbw89: TIL is "today I learned correct" right?

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
seanbw89: There were a ton of typos. I fixed fixed them all.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
lukeschunk: yep 👍

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -1,23 +1,580 @@
...
+          </>
+        )}
+        <CheckBox
+          dataKey="case.otherFeeToWaive"
+          label="Other"
+          name="feesToWaive"
+        />
+        {data.case.otherFeeToWaive && (
+          <>
+            <TextInput label="Explain" dataKey="case.otherFeeToWaive.name" />
```
joeldenning: I think there's a problem here with the data keys.

`case.otherFeeToWaive.name` and `case.otherFeeToWaive`.

The `.` in the data key means "object". But we're trying to store a checkbox's boolean value into `otherFeeToWaive`. And then later we're thinking that `otherFeeToWaive` is an object.

This should be fixed. It matters for both in-memory data storage and for when the user logs in and the data is stored to firebase.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -1,23 +1,580 @@
...
+          name="grossMonthlyIncome"
+          dataKey="person.grossMonthlyIncome"
+        />
+        <small className="web-form-input">
+          (For income that changes from month to month, calculate the annual
+          total and divide by 12 months to list a monthly average.)
+        </small>
+        {data.person.grossMonthlyIncome && (
+          <>
+            <TextInput label="Work" dataKey="person.income.work" />
```
joeldenning: to match the other income data keys, maybe call this `person.income.workIncome`?

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -1,23 +1,580 @@
...
+        <div className="web-form-input">Total:</div>
+        <CheckBox
+          label="I have no income because:"
+          dataKey="person.noGrossMonthlyIncome"
+          name="grossMonthlyIncome"
+        />
+        {data.person.noGrossMonthlyIncome && (
+          <TextArea
+            label="Explain"
+            dataKey="person.noGrossMonthlyIncome.explanation"
```
joeldenning: same here about how `person.noGrossMonthlyIncome` is trying to act as both a boolean and an object.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -1,23 +1,580 @@
...
+          name="monthlyTaxDeductions"
+        />
+      </Section>
+
+      <Section name="7. Monthly expenses">
+        <small className="web-form-input">
+          (Include amounts you pay for yourself and any spouse, children or
+          other dependents in your household)
+        </small>
+        <TextInput label="Rent or mortgage" dataKey="person.income.rent" />
```
joeldenning: We have `person.income.rentalIncome` and then we have `person.income.rent`.

This appears to be in the expenses section. Should it be `person.expenses.rent`? Same goes for all the other data keys in this section

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -1,23 +1,580 @@
...
+
+      <Section name="10. Real Estate">
+        <CheckBox
+          label="I have no real estate"
+          dataKey="person.realEstate"
+          name="realEstate"
+        />
+        <CheckBox
+          label="I have the following real estate"
+          dataKey="person.noRealEstate"
```
joeldenning: Ideally we'd not use two dataKeys for a single boolean value. We don't need to ask two questions and have two checkboxes when they are a single boolean value.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -1,23 +1,580 @@
...
+          name="realEstate"
+        />
+      </Section>
+
+      <Section name="11. Personal Property">
+        <CheckBox
+          label="I have no personal property"
+          dataKey="person.personalProperty"
```
joeldenning: same here about how it would be better to have one datakey and checkbox for a single boolean value, instead of two.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -1,23 +1,580 @@
...
+
+      <Section name="12. Debts Owed">
+        <CheckBox
+          label="I do not owe any debts"
+          dataKey="person.debtsOwed"
+          name="debtsOwed"
+        />
+        <CheckBox
+          label="I do owe the following debts"
+          dataKey="person.noDebtsOwed"
```
joeldenning: we should only have one checkbox, not two

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "cSpell.words": ["TANF", "checkmark", "ocap"]
```
joeldenning: This file should not be in version control. I think the following commands will fix it:

```sh
git rm --cached .vscode/settings.json
git status # it should say that the settings.json file has been deleted
git add .
git commit
```

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.pdf.component.js
```diff
@@ -1,19 +1,625 @@
...
+            "person.middleName"
+          )} ${renderData("person.lastName")}`}
+        </PositionedString>
+        <PositionedString
+          dataKey="person.addressStreet"
+          left={farLeft}
+          top="17.50%"
+        />
+        <PositionedString debugKey="lineAddress" left={farLeft} top="20.90%">
+          {`${renderData("person.addressCity")} ${renderData(
```
joeldenning: Missing comma between the city and state.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.pdf.component.js
```diff
@@ -1,19 +1,625 @@
...
+          dataKey="case.otherFeeToWaive"
+          left="56.3%"
+          top="13.2%"
+        />
+        <PositionedString
+          debugKey="otherFeeToWaiveExplanation"
+          left="64.8%"
+          top="13.3%"
+        >
+          {`${renderData("case.otherFeeToWaiveName")} ( $${renderData(
```
joeldenning: The `$${` is for a dollar sign before the break in the template string? Just making sure the `$$` is intentional

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.pdf.component.js
```diff
@@ -1,19 +1,625 @@
...
+        <PositionedCheckmark
+          dataKey="person.businessInterest"
+          top="28%"
+          left="18%"
+        />
+        <PositionedCheckmark
+          dataKey="person.businessInterest"
+          top="25.7%"
+          left="18%"
+          shouldShow={data.person.businessInterest != true}
```
joeldenning: generally speaking in javascript, it's best to always do a `===` or `!==` check instead of `==` and `!=`.

The reason is that triple equals does not do [type coercion](https://developer.mozilla.org/en-US/docs/Glossary/Type_coercion), but double equals does.

Type coercion is notoriously complex and confusing, so the javascript community usually shuns it. Let's update all of the `!=` and `==` to `!==` and `===`

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.pdf.component.js
seanbw89: Yes that dollar sign is intentional

PR #58, opened 2019-09-27
Fixed scrollable navbar for short screens

HappyViki:


frontend/navbar/menu-items.component.js
```diff
@@ -126,6 +126,7 @@ const css = `
     z-index: 10000;
     left: 0;
     transition: left 0.3s;
+    overflow-y: scroll;
```
joeldenning: `auto` would be better than scroll. The reason is that `scroll` forces a scroll bar to always be there on Windows/Linux, even if it’s unnecessary. Mac’s do not always show the scroll at with overflow scroll.

auto gives you the scrollbar only when necessary. See https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#values for more info.

```suggestion
    overflow-y: auto;
```

PR #60, opened 2019-10-01
WIP: Petition To Expunge Records (Dismissal or Acquittal) Form

benschinn:


frontend/app/forms/petition-for-dismissal-or-acquittal/petition-for-dismissal-or-acquittal.pdf.component.js
```diff
@@ -0,0 +1,5 @@
+import React from "react";
+
+export default function PetitionForDismissalOrAcquittal_Pdf(props) {
+  return <div>PetitionToExpungeRecords_Pdf</div>;
```
lukeschunk: this is a good start - i know this is WIP, but I would just add the `<RenderPage>` component here so people don't get confused if they try to print

lukeschunk:
other than one comment it looks great.

PR #63, opened 2019-10-02
Fix builds for external PRs, closes #62

### LINKED ISSUE #62: Build fails for forks
I think it's because env variables are not made available to PRs in different repos. Not sure though. We should fix this.

jenseng:
Pull requests builds don't deploy to GAE, so there's no need to try to
decrypt the secrets, which doesn't work for external PRs since travis
doesn't provide the encryption key.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,8 @@ deploy:
   on:
     branch: master
 before_install:
-  - openssl aes-256-cbc -K $encrypted_bb8f495d020f_key -iv $encrypted_bb8f495d020f_iv
-    -in app-engine-secret.json.enc -out app-engine-secret.json -d
+  - |
+    if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
+      openssl aes-256-cbc -K $encrypted_bb8f495d020f_key -iv $encrypted_bb8f495d020f_iv \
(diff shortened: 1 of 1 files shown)


jenseng:
Travis build passed 🎉 ... we'll just have to double check once it's merged to be sure it didn't break the deployment somehow 😅

bash

.travis.yml
```diff
@@ -11,5 +11,8 @@ deploy:
   on:
     branch: master
 before_install:
-  - openssl aes-256-cbc -K $encrypted_bb8f495d020f_key -iv $encrypted_bb8f495d020f_iv
-    -in app-engine-secret.json.enc -out app-engine-secret.json -d
+  - |
+    if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
```
lukeschunk: just out of curiosity, where is this env variable set? (i'm assuming it's an env variable)

joeldenning:
The deployment worked! https://travis-ci.org/JustUtahCoders/utahexpungements.org/builds/592756528

.travis.yml
joeldenning: https://docs.travis-ci.com/user/environment-variables/#convenience-variables

PR #64, opened 2019-10-02
Started text input group

seanbw89:
Looking for feedback.

### DIFF
--- a/frontend/app/forms/inputs/text-input-group.component.js
+++ b/frontend/app/forms/inputs/text-input-group.component.js
@@ -16,8 +16,9 @@ export default function TextInputGroup(props) {
           <React.Fragment key={`${input}-${idx}`}>
             {Object.entries(input).map(([key, val]) => (
               <div key={`${key}-${idx}`} className="web-form-input text-input">
-                <label>{key}</label>
+                <label for={`${idx}-${key}`}>{formatLabel(key)}</label>
                 <input
+                  id={`${idx}-${key}`}
                   type="text"
                   value={input[key]}
                   onChange={evt => handleChange(evt, key, idx)}
@@ -41,6 +42,11 @@ export default function TextInputGroup(props) {
     </div>
   );
 
+  function formatLabel(label) {
+    label = label.split(/(?=[A-Z])/).join(" ");
+    return label.charAt(0).toUpperCase() + label.slice(1).toLowerCase();
+  }
+
   function handleChange(evt, key, idx) {
     const values = [...inputList];
     values[idx][key] = evt.target.value;
@@ -48,14 +54,12 @@ export default function TextInputGroup(props) {
   }
   function handleAddGroup(evt) {
     evt.preventDefault();
-    const values = [...inputList];
-    values.push(inputs);
+    const values = [...inputList, inputs];
     setInputList(values);
   }
   function handleRemoveGroup(evt, i) {
     evt.preventDefault();
-    const values = [...inputList];
-    values.splice(i, 1);
+    const values = inputList.filter((item, idx) => idx !== i);
     dataContext.setData(props.dataKey, values);
     setInputList(values);
   }
@@ -66,10 +70,8 @@ export default function TextInputGroup(props) {
 }
 
 const css = `
-	& .text-input {
-		display:flex;
-		flex-direction:column
-	}
-
-	
+  & .text-input {
+    display:flex;
+    flex-direction:column;
+  }
 `;
--- a/frontend/app/forms/inputs/text-input-group.component.js
+++ b/frontend/app/forms/inputs/text-input-group.component.js
@@ -4,10 +4,13 @@ import { DataContainerContext } from "../data-container.component";
 
 export default function TextInputGroup(props) {
   const dataContext = React.useContext(DataContainerContext);
-  var inputs = Object.assign({}, ...props.inputs);
+  var inputs = props.inputs.reduce(
+    (acc, value) => ({ ...acc, [value]: "" }),
+    {}
+  );
   const [inputList, setInputList] = React.useState([inputs]);
   const scope = useCss(css);
-
+  console.log(inputList);
   return (
     <div className="web-form-input" onBlur={evt => handleBlur(evt)} {...scope}>
       <label>{props.groupLabel}</label>
@@ -16,7 +19,7 @@ export default function TextInputGroup(props) {
           <React.Fragment key={`${input}-${idx}`}>
             {Object.entries(input).map(([key, val]) => (
               <div key={`${key}-${idx}`} className="web-form-input text-input">
-                <label for={`${idx}-${key}`}>{formatLabel(key)}</label>
+                <label htmlFor={`${idx}-${key}`}>{formatLabel(key)}</label>
                 <input
                   id={`${idx}-${key}`}
                   type="text"
--- a/frontend/app/forms/inputs/text-input-group.component.js
+++ b/frontend/app/forms/inputs/text-input-group.component.js
@@ -10,7 +10,7 @@ export default function TextInputGroup(props) {
   );
   const [inputList, setInputList] = React.useState([inputs]);
   const scope = useCss(css);
-  console.log(inputList);
+
   return (
     <div className="web-form-input" onBlur={evt => handleBlur(evt)} {...scope}>
       <label>{props.groupLabel}</label>
--- a/frontend/app/forms/inputs/text-input-group.component.js
+++ b/frontend/app/forms/inputs/text-input-group.component.js
@@ -1,0 +1,75 @@
+import React from "react";
+import { useCss } from "kremling";
+import { DataContainerContext } from "../data-container.component";
+
+export default function TextInputGroup(props) {
+  const dataContext = React.useContext(DataContainerContext);
+  var inputs = Object.assign({}, ...props.inputs);
+  const [inputList, setInputList] = React.useState([inputs]);
+  const scope = useCss(css);
+
+  return (
+    <div className="web-form-input" onBlur={evt => handleBlur(evt)} {...scope}>
+      <label>{props.groupLabel}</label>
+      <div>
+        {inputList.map((input, idx) => (
+          <React.Fragment key={`${input}-${idx}`}>
+            {Object.entries(input).map(([key, val]) => (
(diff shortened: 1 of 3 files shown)


frontend/app/forms/inputs/text-input-group.component.js
```diff
@@ -0,0 +1,75 @@
...
+
+  function handleChange(evt, key, idx) {
+    const values = [...inputList];
+    values[idx][key] = evt.target.value;
+    setInputList(values);
+  }
+  function handleAddGroup(evt) {
+    evt.preventDefault();
+    const values = [...inputList];
+    values.push(inputs);
```
lukeschunk: here by pushing, you're mutating state. I would do something like this. `const values = [...inputList, inputs]`

frontend/app/forms/inputs/text-input-group.component.js
```diff
@@ -0,0 +1,75 @@
...
+    evt.preventDefault();
+    const values = [...inputList];
+    values.splice(i, 1);
+    dataContext.setData(props.dataKey, values);
+    setInputList(values);
+  }
+  function handleBlur(e) {
+    e.preventDefault();
+    dataContext.setData(props.dataKey, inputList);
+  }
```
lukeschunk: is the `e.preventDefault()` necessary for all of these? 

Also - and this is personal preference but I would declare these above the return value of the function. Also, I try not to declare too many of these inside of a functional component body.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -27,8 +28,20 @@ const waiveFeeRelationship = [
...
+    EmployersPhone: "",
+    JobTitle: "",
+    HourlyRateOrAnnualSalary: "",
+    HoursPerWeek: ""
+  }
+];
+
 export default function MotionToWaiveFees_Web({ data }) {
   const scope = useCss(css);
+  console.log(data);
```
lukeschunk: remove

frontend/app/forms/inputs/text-input-group.component.js
```diff
@@ -0,0 +1,75 @@
...
+
+  return (
+    <div className="web-form-input" onBlur={evt => handleBlur(evt)} {...scope}>
+      <label>{props.groupLabel}</label>
+      <div>
+        {inputList.map((input, idx) => (
+          <React.Fragment key={`${input}-${idx}`}>
+            {Object.entries(input).map(([key, val]) => (
+              <div key={`${key}-${idx}`} className="web-form-input text-input">
+                <label>{key}</label>
```
joeldenning: For better accessibility, we should connect this label with the input it's labeling. Easiest way might be to put a unique id on the inputs and then do `<label for="the-id">`. Since there are multiple inputs, maybe using the index in the array as part of the id to guarantee uniqueness.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -201,10 +214,22 @@ export default function MotionToWaiveFees_Web({ data }) {
...
             )}
           </>
         )}
+        {data.case.employed &&
+          (data.case.hourlyEmployee ||
+            data.case.salaryEmployee ||
+            data.case.selfEmployed ||
+            data.case.otherEmployment) && (
+            <TextInputGroup
+              groupLabel="Current Employer's"
```
joeldenning: Grammar fix

```suggestion
              groupLabel="Current Employers"
```

frontend/app/forms/inputs/text-input-group.component.js
joeldenning: This is not mutating state -- `values` is a new array that is different from `inputList`. And inputList is the existing state. @seanbw89 what we're referring to here is [this React documentation page](https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly) that explains how modifying state is bad.

You are not doing that, though, so your code is fine. I agree that changing to a one liner might be more clever, but this isn't about mutation of state.

frontend/app/forms/inputs/text-input-group.component.js
```diff
@@ -0,0 +1,75 @@
...
+  function handleAddGroup(evt) {
+    evt.preventDefault();
+    const values = [...inputList];
+    values.push(inputs);
+    setInputList(values);
+  }
+  function handleRemoveGroup(evt, i) {
+    evt.preventDefault();
+    const values = [...inputList];
+    values.splice(i, 1);
```
joeldenning: What you have here is fine, but here's another way that might be easier:
```suggestion
const values = inputList.filter((item, index) => index !== i)
```

frontend/app/forms/inputs/text-input-group.component.js
```diff
@@ -0,0 +1,75 @@
...
+    setInputList(values);
+  }
+  function handleBlur(e) {
+    e.preventDefault();
+    dataContext.setData(props.dataKey, inputList);
+  }
+}
+
+const css = `
+	& .text-input {
```
joeldenning: Please change your IDE to not use tabs. This usually doesn't matter because Prettier can fix it, but template strings are whitespace sensitive so prettier can't change them without altering program behavior.

To do so: 
1. Do this: https://stackoverflow.com/questions/34174207/how-to-change-indentation-in-visual-studio-code
2. F1 -> Convert Indentation to Spaces

frontend/app/forms/inputs/text-input-group.component.js
```diff
@@ -0,0 +1,75 @@
...
+		display:flex;
+		flex-direction:column
```
joeldenning: css requires semicolons. Please add

frontend/app/forms/inputs/text-input-group.component.js
```diff
@@ -0,0 +1,75 @@
+import React from "react";
+import { useCss } from "kremling";
+import { DataContainerContext } from "../data-container.component";
+
+export default function TextInputGroup(props) {
+  const dataContext = React.useContext(DataContainerContext);
+  var inputs = Object.assign({}, ...props.inputs);
```
joeldenning: Do we need this default value? To me it doesn't make sense that someone would use TextInputGroup without passing in the inputs.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
```diff
@@ -27,8 +28,20 @@ const waiveFeeRelationship = [
   }
 ];
 
+const employerInputs = [
```
joeldenning: Why an array of objects instead of just an array of strings? From what I can tell, the empty string values are being completely ignored inside of TextInputGroup.

frontend/app/forms/inputs/text-input-group.component.js
seanbw89: The e.preventDeafult() was necessary because if you don't do it it will bring up the form that prints.

frontend/app/forms/inputs/text-input-group.component.js
seanbw89: From my understanding this is creating a new object. Which is a group of inputs that you can create  more groups from.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
seanbw89: I am using an array of objects here as the key is the input label and the value is the the actual value of the input.

joeldenning: But the user of the TextInputGroup component does not provide values for the input. The values come from the context/dataKey, and always start off as empty.

I still don't think it makes sense for the user of TextInputGroup to pass in an object. It seems like an internal implementation detail of TextInputGroup that is leaking out to the consumers of the component.

joeldenning:
@seanbw89 I'm not sure if we're waiting on each other to both respond on this. I am waiting for you to respond to https://github.com/JustUtahCoders/utahexpungements.org/pull/64#discussion_r330832127. The array of objects with empty string values doesn't make much sense to me.

frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js
seanbw89: Sorry for the late response. I am using the object to group inputs. I have been trying to refactor and I am having a hard time understanding how to add and remove groups of inputs.

joeldenning: Why do the inputs need to be grouped? The example here only has one object inside of the array. Are there going to be cases where there will be two objects inside of the array?

seanbw89: Yes when you add a group it adds another object of group inputs.

joeldenning: Do we have a use case for that?

If not, then let's not add it. The motion to waive fees form doesn't seem to have it as far as I can tell?

If it does, then I would propose the following:

```js
[
  ['firstName', 'lastName']
  ['income', 'profession']
]
```

joeldenning: It's the empty strings that don't need to be there that are bothering me.

seanbw89: Okay that makes more sense. I will get that done.

frontend/app/forms/inputs/text-input-group.component.js
```diff
@@ -4,10 +4,13 @@ import { DataContainerContext } from "../data-container.component";
...
   const dataContext = React.useContext(DataContainerContext);
-  var inputs = Object.assign({}, ...props.inputs);
+  var inputs = props.inputs.reduce(
+    (acc, value) => ({ ...acc, [value]: "" }),
+    {}
+  );
   const [inputList, setInputList] = React.useState([inputs]);
   const scope = useCss(css);
-
+  console.log(inputList);
```
joeldenning: Please remove this console.log

PR #66, opened 2019-10-05
Issue #32 - Petition to Expunge Records (Charges Never Filed)

kthompson55:
Adds the "Charges Never Filed" form to the form bank.

Preview:
Charges Never Filed Form Picture


kthompson55:
I ran `yarn check-format` locally, and got the following:
image

I'm not sure how I got this error.
I ran `yarn prettier 'frontend/**/*' 'backend/**/*'`, and it tells me that:
`[error] No parser could be inferred for file: backend\views\index.ejs`

I haven't touched that file, so I'm confused as to why it's breaking my build.

joeldenning:
I fixed the problem in #67. Please update your branch with latest master and it should fix it.

joeldenning:
Looks like you have other errors besides the `.ejs` file that should not have been checked by prettier at all. I thought it was just because of the `ejs` file but it looks like there were other files, too.

Normally, all of this happens during the pre-commit git hook that husky automatically installs. It is possible that your node_modules or git hooks are messed up which made it so that it didn't run prettier on the files when you committed. Another explanation would be if you did `git commit --no-verify` either manually or through git GUI tool.

To fix, please do the following `yarn prettier 'frontend/**/*' --write`

joeldenning:
👍 i'll take a look at the rest of this tomorrow

frontend/app/forms/order-on-petition-charges-not-filed/order-on-petition-charges-not-filed.pdf.component.js
```diff
@@ -0,0 +1,276 @@
...
+  return (
+    <>
+      <RenderPage url={firstPageUrl}>
+        {/* Personal Information */}
+        <PositionedString debugKey="fullName" left={farLeft} top="14.06%">
+          {`${renderData("person.firstName")} ${renderData(
+            "person.middleName"
+          )} ${renderData("person.lastName")}`}
+        </PositionedString>
+        <PositionedString debugKey="addressStreet" left={farLeft} top="17.46%">
```
joeldenning: An easier way of doing this -- `<PositionedString dataKey="person.addressStreet" left={farLeft} top="17.46%" />`.

See [example](https://github.com/JustUtahCoders/utahexpungements.org/blob/master/frontend/app/forms/petition-for-conviction/petition-for-conviction.pdf.component.js#L33)

Same comment applies to all the other PositionedString components that only are rendering a single data key.

`debugKey` prop = for debugging in devtools because there's more than one dataKey inside it.
`dataKey` prop = please render that data key as the only value of the PositionedString.

frontend/app/forms/order-on-petition-charges-not-filed/order-on-petition-charges-not-filed.web.component.js
```diff
@@ -0,0 +1,105 @@
...
+        )}
+        <TextInput dataKey="case.caseNumber" label={__("case number")} />
+        <TextInput dataKey="case.judgeName" label={__("judge full name")} />
+        <Radio
+          dataKey="case.resolvedBy"
+          label={__("will be resolved by")}
+          options={resolvedByOptions}
+        />
+        {data.case.resolvedBy === "Hearing" && (
+          <TextInput dataKey="case.hearingDate" label={__("hearing date")} />
```
joeldenning: `hearingDate` is too inspecific of a data key, since there can be many hearings for a case. I don't know if you're reusing it from an existing for or whether this is a new data key??

If it's a new dataKey, let's change it to `case.expungementPetitionHearingDate`

frontend/app/forms/order-on-petition-charges-not-filed/order-on-petition-charges-not-filed.web.component.js
```diff
@@ -0,0 +1,105 @@
...
+        )}
+      </Section>
+
+      <Section name="3. Court resolution">
+        <TextInput
+          dataKey="resolution.leaFileNumber"
+          label={__("lea file number")}
+        />
+        <TextInput dataKey="resolution.leaName" label={__("lea name")} />
```
joeldenning: Hmm let's confirm with @tuckersamuelsen about LEA Filing Numbers. I don't know what those are -- they might be the same thing as the case number.

frontend/app/forms/order-on-petition-charges-not-filed/order-on-petition-charges-not-filed.web.component.js
```diff
@@ -0,0 +1,105 @@
...
+        <TextInput dataKey="resolution.date" label={__("resolution date")} />
```
joeldenning: Is this a new dataKey that you created?

Thus far I'm only aware of `person` and `case` as dataKeys. We should put the resolution onto the `case` property, so that the correct resolution is associated with the correct case when there are multiple cases/resolutions.

frontend/app/forms/order-on-petition-charges-not-filed/order-on-petition-charges-not-filed.web.component.js
```diff
@@ -0,0 +1,105 @@
...
+      </Section>
+
+      <Section name="4. Certificate of service">
+        <TextInput
+          dataKey="certificate.prosecutorFullName"
```
joeldenning: Same here with certificate - i'd prefer this to be on the `case` object.

`case.certificate.prosecutorFullName`

Reason is that our data store and UI is based around just the two top-level properties of `case` and `person`. We have two modes in our UI -- one where the data is not persisted and one where it is. You are likely developing in the mode where the data is not persisted because you haven't created a login. I'm worried that introducing these new top-level properties will break the data persistence. There can be multiple cases, multiple certificates, and multiple resolutions. By putting them in separate properties, we'd need a foreign key between them to know which is which.

tuckersamuelsen:
page 3 is filled out by the court. 
 judge name they can fill out but they probably won't know it, so it's OK if that's blank.
The first page, you'll want to check the box next to "the pleadings and other papers of the parties", and not bother with the hearing date field.
The LEA file number is tricky.  It's not the same number as any other form we have in the bank.  I'll have to research where they get that number from.

kthompson55:
@joeldenning I marked some of your comments as resolved, since the fields have been removed altogether, as per the discussion with Tucker on Slack. Copy-pasted from commit message:
* LEA form fields have been removed
* Certificates (page 3) will not be done in form
* Court resolution details (page 2) will not be done in form
* Remove hearing date (page 2) from form, even when "resolved by: *hearing" is selected

The LEA fields are still in limbo until we get more information from @tuckersamuelsen

frontend/app/forms/order-on-petition-charges-not-filed/order-on-petition-charges-not-filed.pdf.component.js
```diff
@@ -0,0 +1,147 @@
...
+        <PositionedString
+          dataKey="person.email"
+          debugKey="email"
+          left={farLeft}
+          top="28.16%"
+        />
+
+        {/* Case information */}
+        <PositionedCheckmark
+          debugKey="courtTypeDistrict"
```
joeldenning: debugKey is unneeded if you have dataKey on PositionedString.

Doesn't hurt though -- it really is just for debugigng in the devtools. The data key for PositionedString also shows up in dev tools.

joeldenning:
This will be deployed to utahexpungements.org in https://travis-ci.org/JustUtahCoders/utahexpungements.org/builds/594329964

Issue #35, opened 2019-07-17
Form - Reply to Victim's or Prosecutor's Statement


joeldenning:
https://www.utcourts.gov/howto/expunge/docs/03_Petitioner_Reply.pdf

PR #72, opened 2019-10-12
Issue #35 - Reply to Victim's or Prosecutor's Statement

kthompson55:
The TextArea component has a max input length. Even at the maximum length, the PDF form version won't overrun the first page's available space for the statement of disagreement.

image

Page 2 has a signature area, so I didn't write an text inputs for that, and Page 3 is a Certificate of Service page that I believe gets filled out at the courthouse.


frontend/app/forms/reply-to-statement/reply-to-statement.pdf.component.js
```diff
@@ -0,0 +1,153 @@
...
+          dataKey="person.addressStreet"
+          left={farLeft}
+          top="17.46%"
+        />
+        <PositionedString
+          debugKey="addressStreetCityZip"
+          left={farLeft}
+          top="20.86%"
+        >
+          {`${renderData("person.addressCity")} ${renderData(
```
joeldenning: Might be better to have a comma between the city and state?

frontend/app/forms/reply-to-statement/reply-to-statement.pdf.component.js
```diff
@@ -0,0 +1,153 @@
...
+
+      <RenderPage url={secondPageUrl}>
+        <PositionedString
+          dataKey="case.disagreementReason"
+          left={disagreementLeft}
+          top="28.0%"
+          style={{
+            width: "71%",
+            overflowWrap: "anywhere",
+            lineHeight: "31px"
```
joeldenning: Hmm which browsers have you tested this on? I'm a bit worried that the 8.5x11" print preview might display 31px differently in different browsers.

frontend/app/forms/reply-to-statement/reply-to-statement.web.component.js
```diff
@@ -0,0 +1,78 @@
...
+        <TextInput dataKey="case.judgeName" label={__("judge full name")} />
+      </Section>
+
+      <Section name="3. Statement response">
+        <TextArea
+          dataKey="case.opposingStatement"
+          label={__("statement made by opposition")}
+        />
+        <TextArea
+          dataKey="case.disagreementReason"
```
joeldenning: Let's clarify this dataKey's meaning
```suggestion
          dataKey="case.opposingStatementDisagreementReason"
```

frontend/app/forms/reply-to-statement/reply-to-statement.pdf.component.js
kthompson55: I was developing in FireFox. I will check Chrome, Edge, and IE11 and get back to you on this.

kthompson55: The app doesn't even load in IE11, so I don't have any pictures for it, but here's a comparison of Chrome, Edge, and FireFox:
image

Unfortunately, having a bunch of really long words causes more frequent line breaks, so there's an overflow of text from the first page area. But the text does align properly with 31px as a lineHeight in all browsers.

I don't think many of our users will be rolling their hand over their keyboard like I did just to get a long input, but we might need to account for unexpected line breaks, especially if the user presses "enter" a lot.

joeldenning: 👍 thanks for doing that. I think that we don't need to worry about overflowing beyond the allotted area. And am happy to see that it works in all browsers.

joeldenning: (except IE which we don't support)

kthompson55: Are we good to resolve this conversation, then?

Issue #61, opened 2019-10-02
rem css issue


joeldenning:
In some browsers the app looks huge sometimes. It is because we have one rem set to one pixel. We should change it to just leave the browser default in place of one rem set to 16px.

This will requiring updating all css in all components to use the new rem size.

HappyViki:
Is this still an issue? Can you share which browsers, pages, and screenshots of where it's happening?

I want to work on this.

joeldenning:
Yes -- on Firefox and Chrome intermittently the browser is completely the wrong size. Most recently @kthompson55 was seeing this on his machine.

To fix it, delete this line of code:

https://github.com/JustUtahCoders/utahexpungements.org/blob/8e040d8b8e7f9a4c50d8bbb506a819257976ade1/frontend/styleguide.js#L29

That will change the font size for the root element, which determines the size of the `rem` unit. Next go through all the css in all of the app -- wherever `rem` is used you need to multiple by 16 since the default size for `rem` if left unspecified is `16px`. Check out the `rem` unit [on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/length) where it describes how 16px is the common unit for rems.

Cuthbert20:
Hi, I know the issue was already fixed but, I was thinking what if we just set font-size to a percent so assuming 100% is 16px and we wanted 1px to be our default font-size so we can use rem for margin or padding to make media queries easier we could set font-size to 1 divided by 16 (6.25%). That way we can still use rem's for styling and the user could change the size of the font in their browser.  
I just thought it might be a faster fix but, I could be wrong.  :)

joeldenning:
> what if we just set font-size to a percent so assuming 100% is 16px and we wanted 1px to be our default font-size so we can use rem for margin or padding to make media queries easier we could set font-size to 1 divided by 16 (6.25%)

The problem is that browsers don't like it when you set the root font size so small. Chrome sometimes refuses to respect that font-size and then the app looks weird. The code currently already does set the font size that small, and it doesn't work in all browsers.

joeldenning:
This was fixed in #81

PR #81, opened 2019-10-23
Fixed font-size rem default 16px (#61)

HappyViki:
Root font-size is 16px, rest font-size are rem. I used this [ref](http://www.standardista.com/px-to-rem-conversion-if-root-font-size-is-16px/) to convert.
All the rems that were not font-size have been changed to px.


frontend/about/about.component.js
```diff
@@ -127,10 +127,10 @@ export const css = `
 
   & .main-content {
     width: 75vw;
-    max-width: 936rem;
-    margin: ${rawNavbarHeight + 32}rem auto;
-    padding: 32rem;
-    border: 1rem solid ${darkGray};
+    max-width: 936px;
```
joeldenning: Oh no! I think there was a miscommunication. We **still want to use rems**, but just based on 1rem equaling 16px Sorry for the wasted work! All of your changes will need to be adjusted to be rems.

```suggestion
    max-width: 58.5rem;
```

joeldenning:
Please merge the master branch into your branch, which will fix the build (see related #85)

HappyViki:
I learned how to [copy from one branch to another](https://stackoverflow.com/questions/35174544/git-how-to-copy-contents-of-one-branch-to-other-branch#answer-35177459), woot! I learned about `font-size: 62.5%` from [this](https://medium.com/code-better/css-units-for-font-size-px-em-rem-79f7e592bb97) article.

Anyway, here's what I did:
```
:root { font-size: 62.5%; ... }
body { font-size: 1.6rem; ... }
```
And changed all previous `rem`s to what it was before divided by ten, example:
```
& button, & a.button {
  font-size: 18rem -> 1.8rem;
  border-radius: 6rem -> 0.6rem;
  ...
}
```
`font-size: 62.5%` in root makes it easier to translate default `px` -> `rem`, as you can see in the above example. If `font-size: 100%` in root, then you'd have to do something like `font-size: 18rem -> 11.25rem`, aka multiply by 62.5%.

Body font-size used to be 16px. I treated it like the other rems. You could say I did `16px` -> `16rem` -> `1.6rem`. If I removed that line, body `font-size` would be `16px * 62.5%` by default.

I hope I explained well. I did this mainly to learn from what I did today.

PR #102, opened 2019-10-29
Parsing disposition date. Resolves #94.

### LINKED ISSUE #94: Parse disposition date from PDF

jamesschlader:
See #94


backend/apis/parse-docket-pdf.utils.js
```diff
@@ -15,6 +15,9 @@ exports.parsePdfText = function parsePdfText(text) {
     }
   }
 
+  console.log(`lines`, lines);
```
joeldenning: Generally speaking, we should not commit/push console.logs. Reasons:

- It's noise in the terminal for everyone else
- It's noise in our production logs that we'll have to sort through
- Putting sensitive data (like the contents of a pdf with court info in it) into logs changes the security requirements for logs in a way that is usually undesireable.

PR #107, opened 2019-10-31
Parse fines

jamesschlader:
This pull does three things: 
(1) abstracted the walkSection function so that it could be used by any section (which seems to have been written for that purpose to begin with)
(2) edited the parseCharges function by defining its walk function and then using the walkSection function referencing that walk function. This may not have been necessary, but it could allow for pulling the walk function out of the body of the parseCharges function.
(3) fixed a bug that was adding a new CASE NUMBER section whenever that phrase occurs (at the top of every page). Any section whose content cross over to a separate page was being lost due to the bug.


backend/apis/parse-docket-pdf.utils.js
```diff
@@ -1,13 +1,20 @@
-const sectionHeaderRegex = /^([A-Z][A-Z ]+)/;
+const sectionHeaderRegex = /^([A-Z][A-Z ]+)/g;
```
joeldenning: Why is the global flag needed for this regex? We're applying it just to one line at a time, right? And only care about the first match.

backend/apis/parse-docket-pdf.utils.js
```diff
@@ -72,15 +80,25 @@ function parseCharges(lines, sections) {
...
+  });
+
+  return accounts;
+}
+
 function walkSection(lines, sections, name, walk) {
-  const sectionIndex = sections.findIndex(
-    section => section.name === "CHARGES"
-  );
+  const sectionIndex = sections.findIndex(section => section.name === name);
```
joeldenning: 👍 good catch

backend/apis/parse-docket-pdf.utils.js
```diff
@@ -1,13 +1,20 @@
...
 exports.parsePdfText = function parsePdfText(text) {
   const lines = text.split("\n");
   const sections = [];
+  let gotCase = 0;
 
   for (let i = 0; i < lines.length; i++) {
     const line = lines[i];
     const match = sectionHeaderRegex.exec(line);
-    if (match) {
+    if (match && match[1].trim() !== "CASE NUMBER") {
```
joeldenning: What if we just avoided all duplicate sections? Do you know of any cases where we need any duplicate sections?

```js
for (let i = 0; i < lines.length; i++) {
  const line = lines[i];
  const match = sectionHeaderRegex.exec(line);
  if (match) {
    const name = match[1].trim();
    if (!sections.some(s => s.name === name) {
      sections.push({lineNumber: i, name})
    }
  }
}
```

backend/apis/parse-docket-pdf.utils.js
```diff
@@ -72,15 +80,25 @@ function parseCharges(lines, sections) {
...
+  walkSection(lines, sections, "CHARGES", addDispositionToCharges);
 
   return charges;
 }
 
+function parseAccountSummary(lines, sections) {
+  const accounts = [];
+
+  walkSection(lines, sections, "ACCOUNT SUMMARY", line => {
+    accounts.push(line);
```
joeldenning: Ideally we'd do better than just pushing the raw text into an array, but only extract the relevant values from the account summary section. Is that something you're planning on working on in a future PR?

backend/apis/parse-docket-pdf.utils.js
jamesschlader: The global flag gives access to the lastIndex property. This could prove useful for many lines that satisfy the regex but hold additional info. Some lines look like this: 
TRUST TOTALS Trust Due: 1,050.00
The TRUST TOTALS matches the regex but we'd want to slice what comes next and put that somewhere. So the lastIndex property could be useful for that kind of case.

backend/apis/parse-docket-pdf.utils.js
jamesschlader: The file I was using for testing is pretty big (Evans 141100464) and had 20 some odd pages. The only repeating section was the CASE NUMBER. After implementing the code, I only got back the top line sections ending in PROCEEDINGS. I believe that was the desired result. No other sections  heading repeat so there are no other section duplicates. Worth writing a test to prove it?

backend/apis/parse-docket-pdf.utils.js
joeldenning: You're not storing the lastIndex property in the section objects, though. Additionally, the `TRUST TOTALS` line is not.a section header and does not match this regular expression. Note how this regular expression requires the first letter of the line to be a capital letter, but that the TRUST TOTALS line in [this example docket](https://drive.google.com/drive/u/0/folders/1cW6RM4jLDeiLeSy7emI8_xdf6HhpfBhG) has space characters at the beginning of the line.

`sectionHeaderRegex` is meant to find the top-level sections (like `ACCOUNT SUMMARY`). It is not meant to parse information out of lines within those top level sections (like `TRUST TOTALS`).

backend/apis/parse-docket-pdf.utils.js
jamesschlader: Yeppers. This is why I was asking about a general parsing strategy. It turns out that if you look at the weeds of the PROCEEDINGS, any payment or fine balance = 0 should appear there as well. We could get a quick answer to "Does this person have a 0 account balance on all accounts?" by parsing through the ACCOUNT SUMMARY section but my fear is that we'll need to parse the entire PROCEEDINGS section anyway, so why not answer that question there? Or are we planning on having multiple functions passing over a fully parsed document perhaps doing double duty? Or do we just clean up the data here, saving everything, and test it's qualification status somewhere else? Or what? These questions need to be answered for the team before any further time is wasted on a more mature parse of the account. Hence, I just sent the whole bloody thing back to the client for now.

backend/apis/parse-docket-pdf.utils.js
joeldenning: If case number is the only section that repeats, then okay we can only check for that section 👍 

Let's clean up your code a bit though. For example, we don't need to copy/paste the `sections.push()` or `match[1].trim()` statements.

Here's a suggested implementation that might be a bit easier to follow:

```js
let foundCaseNumberSection = false

for (let i = 0; i < lines.length; i++) {
  const line = lines[i];
  const match = sectionHeaderRegex.exec(line);
  if (match) {
    addSection(match[1], i)
  }
}

function addSection(name, lineNumber) {
  const shouldAdd = name !== 'CASE NUMBER' || !foundCaseNumberSection
  if (name === 'CASE NUMBER') {
    foundCaseNumberSection = true
  }
  if (shouldAdd) {
    sections.push({name, lineNumber})
  }
}
```

backend/apis/parse-docket-pdf.utils.js
joeldenning: Let's figure out the answers to those questions in future PRs. I'm okay with merging in what you have here for the account summary and then iterating on it in future PRs.

backend/apis/parse-docket-pdf.utils.js
jamesschlader: Do you want me to update the code with this suggestion (thanks for that) and remove the g flag from the regex?

joeldenning: yep, remove the `g` flag from the regex since it's not being used or necessary (at least for now)

Issue #78, opened 2019-10-23
Pardon Application Form


joeldenning:
https://bop.utah.gov/images/pdf/PardonApplication.pdf

scotttball:
@joeldenning this one is going to take me a day longer. I have to make some new data types to handle the new fields. :(

joeldenning:
👍 no worries. Thanks again for your help! I will share with you over slack the next initiative within utahexpungements.org after you get that one finished.

PR #109, opened 2019-11-01
WIP: Application for Pardon PDF/Web/pngs

scotttball:
WIP'n this. I have to create new data points to take in all the new data that this application needs. 

#### Web Component
- [x] new fields spouse and children
- [ ] new fields for employment history
- [ ] new fields references
- [ ] new fields Convictions Sought To Be Pardoned
- [ ] new fields Other Arrests or Convictions
- [ ] new fields Incarceration History
- [ ] new fields Incarceration Disciplinary History
#### Pdf Component
- [ ] styling css

Finishes #78


frontend/app/forms/application-for-pardon/application-for-pardon.web.component.js
```diff
@@ -0,0 +1,202 @@
...
+              label={__("previously please explain")}
+            />
+          )}
+        </Section>
+        <Section name="3. Family History">
+          <p className="web-form-input">
+            If married, please provide the following information for your spouse
+          </p>
+          <TextInput
+            dataKey="spouse.firstName"
```
joeldenning: Is `spouse.firstName` a new data key?

The `.` in data keys represents a sub object of a collection in firebase. We have two modes for the data entry - first is just in-memory store and the second is with a login via firebase.

Right now we only have two collections - `case` and `person`. We allow users to switch which case and which person (in the case they're a lawyer they have multiple clients) they're working on. By introducing a new top-level collection `spouse`, you're (probably inadvertently) indicating that spouse is also something that could be swapped out independent of the person. Which isn't something that we have a UI for, nor is it a feature that we want.

Instead, this should be `person.spouseFirstName` or something similar, just tacked on as a property on the `person` collection.

frontend/app/forms/application-for-pardon/application-for-pardon.web.component.js
```diff
@@ -0,0 +1,202 @@
...
+          <TextInput dataKey="spouse.addressStreet" label={__("street")} />
+          <TextInput dataKey="spouse.addressCity" label={__("city")} />
+          <TextInput dataKey="spouse.addressState" label={__("state")} />
+          <TextInput dataKey="spouse.addressZip" label={__("zip")} />
+          <TextInput dataKey="spouse.homePhone" label={__("home phone")} />
+          <TextInput dataKey="spouse.dayPhone" label={__("day phone")} />
+          <TextInput dataKey="spouse.email" label={__("email address")} />
+          // children of person
+        </Section>
+        <Section name="4. Employment History">
```
joeldenning: See related https://github.com/JustUtahCoders/utahexpungements.org/blob/master/frontend/app/forms/motion-to-waive-fees/motion-to-waive-fees.web.component.js#L229 when you start working on this. There might be some dataKeys that can be reused there.

frontend/app/forms/form-common-options/form-common-options.js
```diff
@@ -59,6 +59,17 @@ export const courtTypeOptions = [
   { label: "Justice Court", value: "Justice" }
 ];
 
+export const previouslyAppliedOptions = [
```
joeldenning: I think this should maybe just be a variable inside of the application for pardon web component. I don't know of another form that asks whether the petitioner has previously applied or not.

Additionally, the label and value here are just a Yes/No, which isn't really a list of common options that is specific to this "previously applied" question.

frontend/app/forms/application-for-pardon/application-for-pardon.web.component.js
scotttball: Yes, it is a new data key. 

Yep, sounds good. I made this commit when I didn't know what structure our data was.

PR #116, opened 2019-11-06
Added pdf text extractor option to docket-pdf

HappyViki:
Fetches and displays either parsed JSON or whole text from file, depending on which radio button is checked.
To temporarily fix the bug my code introduced: whether or not the `fetch` returned data, it will `setIsProcessing(false)`.


HappyViki:
I think processed PDF format should be default. What I want to do is instead of checking if I want the JSON: checking if I want the plain text. What do you think?

frontend/app/docket-pdf/docket-pdf.component.js
```diff
@@ -3,31 +3,37 @@ import React from "react";
...
             } else {
               throw Error(
                 `Server responded with error ${JSON.stringify(data.error)}`
               );
             }
-          })
-        )
+          };
+          if (processJson) {
+            r.json().then(gotProcessed);
```
joeldenning:
```suggestion
            return r.json().then(gotProcessed);
```

frontend/app/docket-pdf/docket-pdf.component.js
```diff
@@ -3,31 +3,37 @@ import React from "react";
...
+          } else {
+            r.text().then(gotProcessed);
```
joeldenning:
```suggestion
            return r.text().then(gotProcessed);
```

Issue #95, opened 2019-10-29
Separate offense name from severity


joeldenning:

Cuthbert20:
I am thinking of splitting @ 1st || 2nd || 3rd || Class. My only concern is if there is an offense name that includes the word "class" in it. 
I am going try and find a list of offense names and see if there are any that include the word class.

joeldenning:
I think one implementation approach could be to find the last word that is all caps and then split the string based on that.

Cuthbert20:
yeah that's a great idea.
Thanks!

tuckersamuelsen:
I just pulled up about a bunch of old court cases, and I have not found any counter examples to the option @joeldenning presented.  I think we can safely say that the offense description will ALWAYS be all caps, and the severity will ALWAYS be first letter capitalized, otherwise lowercase.

Issue #100, opened 2019-10-29
Alter docket pdf API so that it indicates if it's missing data, or if it was unable to parse the pdf, etc.


joeldenning:
We want to know how confident we are about whether the information we have is parseable and enough

PR #118, opened 2019-11-07
#95 Seperate offense name from severity

Cuthbert20:
Separated Description into two different values (offenseName & severity).


backend/apis/parse-docket-pdf.utils.js
```diff
@@ -84,8 +84,16 @@ function parseCharges(lines, sections) {
   }
 
   walkSection(lines, sections, "CHARGES", addDispositionToCharges);
+  let chargesWithSeverity = charges.map(elm => {
+    let index = elm.description.search(/.[A-Z][a-z]|.\d[a-z]/);
+    let offenseName = elm.description.slice(0, index);
+    let severity = elm.description.slice(index + 1);
+    elm = { ...elm, offenseName, severity };
```
scotttball: Great work. I would suggest just creating a new object and then spreading elm to it instead of mutating the `elm` object with a new one.

ex:
```javascript
const { description, ..rest } = elm;
const index = description.search(/.[A-Z][a-z]|.\d[a-z]/);
const offenseName = description.slice(0, index);
const severity = description.slice(index + 1);
const newElm = { ...rest, offenseName, severity };
return newElm

```

also, not entirely sure what elm is, is it a charge?

backend/apis/parse-docket-pdf.utils.js
```diff
@@ -84,8 +84,16 @@ function parseCharges(lines, sections) {
   }
 
   walkSection(lines, sections, "CHARGES", addDispositionToCharges);
+  let chargesWithSeverity = charges.map(elm => {
```
joeldenning: In general, a good rule of thumb is to start variables out as `const`s and then switch them to `let`s as needed. Also, renaming the `elm` variable makes the rest of the code easier to follow:

```suggestion
  const chargesWithSeverity = charges.map(charge => {
```

backend/apis/parse-docket-pdf.utils.js
joeldenning: 👍 agreed

backend/apis/normal-docket-pdf.test.js
```diff
@@ -17,13 +17,15 @@ describe("normal docket pdf file", () => {
     let actual1 = parsedText.charges[1];
 
     expect(actual0.statute).toBe("53-3-202(1)(A)");
-    expect(actual0.description).toBe("NO VALID LICENSE Misdemeanor");
+    expect(actual0.offenseName).toBe("NO VALID LICENSE");
```
joeldenning: From https://github.com/JustUtahCoders/utahexpungements.org/blob/ceac1f1e418a35552a3b53f0d9412e0a864cd224/backend/apis/docket-pdf-test-files/normal-docket-pdf.js#L10-L11

This should be `NO VALID LICENSE - EXPIRED`

backend/apis/normal-docket-pdf.test.js
```diff
@@ -17,13 +17,15 @@ describe("normal docket pdf file", () => {
...
+    expect(actual0.severity).toBe("Misdemeanor");
```
joeldenning: This should be `Class C Misdemeanor`

backend/apis/normal-docket-pdf.test.js
```diff
@@ -17,13 +17,15 @@ describe("normal docket pdf file", () => {
...
     expect(actual0.disposition).toBe("Dismissed (w/o prej)");
     expect(actual0.dispositionDate).toBe("Month 00, 0000");
 
     expect(actual1.statute).toBe("41-6A-601");
-    expect(actual1.description).toBe(
-      "SPEEDING 70 in a 65 Class C Misdemeanor Offense Date: Month 00, 0000"
+    expect(actual1.offenseName).toBe("SPEEDING 70 in a 65");
```
joeldenning: 👍

backend/apis/normal-docket-pdf.test.js
```diff
@@ -17,13 +17,15 @@ describe("normal docket pdf file", () => {
...
+    expect(actual1.severity).toBe(
+      "Class C Misdemeanor Offense Date: Month 00, 0000"
```
joeldenning:
```suggestion
      "Class C Misdemeanor
```

backend/apis/parse-docket-pdf.utils.js
```diff
@@ -84,8 +84,16 @@ function parseCharges(lines, sections) {
...
+    let index = elm.description.search(/.[A-Z][a-z]|.\d[a-z]/);
```
joeldenning: I spent some time looking at this regex. I don't think it correctly finds the parse index. The suggestions I made in the test file shows examples of how the regex is failing to find the correct index. It's ironic that the one test case we have is one where the all caps vs lower case rule (discussed in https://github.com/JustUtahCoders/utahexpungements.org/issues/95) doesn't actually work 😭 

Here's what I think we should do:
1. Add in another test pdf file that *does* have the all caps vs lowercase distinction for offenseName and severity.
2. Change your code to look for the last uppercase word. Something like this:
```js
const words = charge.description.split(/\s/)
const splitIndex = words.findIndex(word => word.toUpperCase() !== word)
const offenseName = words.slice(0, splitIndex).join(' ')
const severity = words.slice(splitIndex).join(' ')
```
3. Brainstorm what we can do about the weird `SPEEDING 70 in a 65` offense name that has some lowercase letters in it even though offense names usually don't have lowercase letters. This offense name seems to be an exceptional case to me, not the norm. Let's try to find as many other exceptional cases as we can, too, in the test data. But if this is the only one, we might be able to get away with just treating it special.

backend/apis/parse-docket-pdf.utils.js
Cuthbert20: Sweet I like that better! thanks @scotttball I will make changes

backend/apis/parse-docket-pdf.utils.js
Cuthbert20: Ok yeah that makes sense, I should have made them all const's and naming makes sense to make it easier to read. thanks

backend/apis/normal-docket-pdf.test.js
Cuthbert20: ok yeah that makes sense, I just split description up on the current tests. I should have made my own tests. thanks I will work on the tests

backend/apis/parse-docket-pdf.utils.js
Cuthbert20: ha yeah it is ironic. I will go through all the pdfs and see if I can find any other exceptional cases. If I find more I will let you know.  
I will work on resolving the issues. Thanks for the feedback @joeldenning !

Cuthbert20: I was going through the pdf's and it looks like if the offenseName starts with SPEEDING followed by a number we get "in a" followed by another number. I would assume (I know its bad to assume) that both numbers before "in a" and after will always be atleast 2 digits long ie SPEEDING 60 in a 30 and that might be a way to deal with these special cases.

backend/apis/normal-docket-pdf.test.js
Cuthbert20: Yes I agree it should be I am looking at the pdf right now.  I think the issue is this charge has 3 instances of " - " where normally there are only 2 from what I have seen so far. So we are getting NO VALID LICENSE and then we get back Misdemeanor because it is on the second line
below is what is on the pdf. and we are losing EXPIRED Class C on the split. I could be wrong though.
 Charge 1 - 53-3-202(1)(A) - NO VALID LICENSE - EXPIRED Class C 
         Misdemeanor

backend/apis/normal-docket-pdf.test.js
Cuthbert20: totally agree

backend/apis/parse-docket-pdf.utils.js
```diff
@@ -85,7 +85,17 @@ function parseCharges(lines, sections) {
 
   walkSection(lines, sections, "CHARGES", addDispositionToCharges);
 
-  return charges;
+  const chargesWithSeverity = charges.map(charge => {
+    const { description, rest } = charge;
```
joeldenning:
```suggestion
    const { description, ...rest } = charge;
```

Using rest args isn't even necessary here, though. See my comment on line 94

backend/apis/parse-docket-pdf.utils.js
```diff
@@ -85,7 +85,17 @@ function parseCharges(lines, sections) {
...
+    const words = description.split(/\s/);
+    const splitIndex = words.findIndex(word => word.toUpperCase() !== word);
+    const offenseName = words.slice(0, splitIndex).join(" ");
+    const severity = words.slice(splitIndex).join(" ");
+    const charges = { ...rest, offenseName, severity };
```
joeldenning:
```suggestion
    const charges = { ...charge, offenseName, severity };
```

The offenseName and severity will override anything in the charge object, since they are after the `...charge` inside of the object.

Cuthbert20: I noticed that right after I pushed. I apologize I fixed it right after I pushed. I add and commuted I just didn’t push when I get home I will push it up.

Cuthbert20: Oh ok I see now what your saying. I will adjust and remove rest

Cuthbert20: The issue I am having when I remove the rest arg is that when I spread charges along with offenseName & severity charge still includes description. Unless I am doing it wrong I will push up my code right now and maybe you can tell me if I misunderstood.
Thanks for the feedback!

joeldenning: Ah I see. I didn't realize that we were trying to remove properties from the charge. Yeah let's go back to using the rest args like you have above ^. 👍

backend/apis/normal-docket-pdf.test.js
joeldenning: the dashes aren't how we should be splitting the description string, though - it should be based on uppercase and lowercase words. Right?

joeldenning:
This looks good except that the tests are failing. See https://travis-ci.org/JustUtahCoders/utahexpungements.org/builds/610975282?utm_source=github_status&utm_medium=notification. We just need to handle the `Driving in 70 in a 60` offense name

Cuthbert20:
Hey @joeldenning I was wondering if there was anything you wanted me to change after your review. If not I can start working on issue #100 If there are changes that need to be made I will get on them.

PR #119, opened 2019-11-07
feature: eslint added and fixed a few eslint errors

### LINKED ISSUE #117: Add eslint
For frontend, we can use https://github.com/joeldenning/eslint-config-react-important-stuff. For backend we can maybe reuse the same config, or use https://github.com/joeldenning/eslint-config-node-important-stuff

scotttball:
Closes #117 

* added eslint
* added `joeldenning/eslint-config-react-important-stuff`
* fixed up .eslintrc
* husky/lint-staged for frontend
* added `yarn lint`

_currently only eslint for frontend (need to configure for ejs and node to further checks for backend)_

### DIFF
--- a/frontend/app/forms/inputs/text-area.component.js
+++ b/frontend/app/forms/inputs/text-area.component.js
@@ -21,7 +21,7 @@ export default function TextArea(props) {
 
   React.useEffect(() => {
     setValue(get(dataContext.data, props.dataKey, ""));
-  }, [dataContext]);
+  }, [dataContext, props.dataKey]);
 
   return (
     <div {...scope}>
--- a/frontend/app/overview/vocabulary/defined-term.component.js
+++ b/frontend/app/overview/vocabulary/defined-term.component.js
@@ -10,7 +10,26 @@ export default function DefinedTerm(props) {
   const containerRef = useRef(null);
   const alwaysLink =
     useContext(DefinedTermContext).alwaysLink || props.alwaysLink;
+
+  const openTooltip = () => {
+    setTooltipOpen(true);
+  };
+  const closeTooltip = () => {
+    setTooltipOpen(false);
+  };
+
   useEffect(() => {
+    const maybeCloseTooltip = evt => {
+      if (
+        containerRef.current === evt.target ||
+        containerRef.current.contains(evt.target)
+      ) {
+        // clicked on or inside of the popup
+        return;
+      } else {
+        openTooltip();
+      }
+    };
     if (tooltipOpen) {
       document.addEventListener("click", maybeCloseTooltip);
     } else {
@@ -20,7 +39,7 @@ export default function DefinedTerm(props) {
     return () => {
       document.removeEventListener("click", maybeCloseTooltip);
     };
-  }, [maybeCloseTooltip, tooltipOpen]);
+  }, [tooltipOpen]);
 
   const children = props.children || __(`def word - ${props.term}`);
   const Definition = definitionComponents[props.term];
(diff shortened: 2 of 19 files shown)


scotttball:
I still need to do some manual testing on this. I've had issues with some other projects with eslint rules and hooks. I've ran into some infinite rendering issues with the props that eslint recommends to observe. Sometimes it's not always correct. I'll check it out tomorrow. 

Please comment if there are any changes that should be made to this PR.

frontend/app/forms/inputs/text-area.component.js
```diff
@@ -21,7 +21,7 @@ export default function TextArea(props) {
 
   React.useEffect(() => {
     setValue(get(dataContext.data, props.dataKey, ""));
-  }, [dataContext]);
+  }, [dataContext, props.dataKey]);
```
joeldenning: Wow yes all these effect dependency array errors is a great reason why we should have had linting all along

frontend/app/overview/vocabulary/defined-term.component.js
```diff
@@ -10,7 +10,26 @@ export default function DefinedTerm(props) {
   const containerRef = useRef(null);
   const alwaysLink =
     useContext(DefinedTermContext).alwaysLink || props.alwaysLink;
+
+  const openTooltip = () => {
```
joeldenning: Why did these functions get moved, and why switched to consts + arrow function?

frontend/app/user/cases-list.component.js
```diff
@@ -35,6 +35,7 @@ export default class CasesList extends React.Component {
                     </button>
                   )}
                   <button className="danger small">Delete</button>
+                  {/* delete currently does nothing? */}
```
joeldenning: Ha good point, yeah we should fix that.

frontend/app/utils/accordion.component.js
```diff
@@ -30,14 +30,16 @@ export default class Accordion extends React.Component {
 
             return (
               <div key={index}>
-                <a
+                <div
```
joeldenning: Might be better to use `<button className="unstyled">`. It's easier to get accessibility right by using `<button>` than by `<div role="button">`

See https://github.com/JustUtahCoders/utahexpungements.org/blob/92ec0e5ca1433231e0d46037175031192d4942eb/frontend/styleguide.js#L146

scotttball: Agreed! I’ll make the change later today.

frontend/app/overview/vocabulary/defined-term.component.js
scotttball: No reason. I think I was just fooling around a bit. I’ll change the toggles back. There’s an eslint warning for the event function and that’s why I moved it in the useEffect.

scotttball: I needed the `openTooltip` in the useEffect so that's why I moved them up. Would you like me change them back to declarations?

joeldenning: if they are declarations, they are hoisted and can be used anywhere. This is actually why I usually use `function() {}` instead of `const thing = () => {}`.

I don't really mind if you write your own code to use `const thing = () => {}`, but yeah let's not change all of the existing components to it.

joeldenning:
Just one remaining comment, then let's merge https://github.com/JustUtahCoders/utahexpungements.org/pull/119#discussion_r343513803

frontend/app/overview/vocabulary/defined-term.component.js
scotttball: Ok! Sounds good.

PR #122, opened 2019-11-09
Motion To Reduce Conviction 1st Page

benschinn:
Below is a screenshot of some new inputs regarding:
* _plaintiff vs. defendant_, _conviction degree_
* _requested reduced conviction degree_

Screen Shot 2019-11-09 at 12 33 59 AM

----

Screen Shot 2019-11-09 at 12 35 14 AM

* **A** and **B** are values of the aforementioned inputs printed to the document. 
* **C** I wasn't sure if this section is filled out bu the petitioner or not. @tuckersamuelsen do you know?


frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.pdf.component.js
```diff
@@ -1,5 +1,127 @@
...
+        {requestedConvictionDegreeInformation(
+          data.case.convictionDegree,
+          data.case.reducedConvictionDegree
+        )}
+      </RenderPage>
+    </>
+  );
+}
+
+function requestedConvictionDegreeInformation(degA, degB) {
```
benschinn: this render function handles the (1) section [here](https://www.utcourts.gov/howto/criminallaw/docs/01_Motion_to_Reduce_Conviction.pdf). The positioning of the values should change depending on whether the petitioning for one or two degree reduction. I'd love feedback on this logic and whether there can be improvements.

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.web.component.js
```diff
@@ -1,5 +1,83 @@
...
+  DistrictCourtList
+} from "../form-common-options/form-common-options";
 
-export default function MotionToReduceConviction_Web() {
-  return <div>Web component</div>;
+export default function MotionToReduceConviction_Web({ data }) {
+  return (
+    <FormThatPrints>
+      <Section name="1. Personal Information">
+        <TextInput dataKey="person.firstName" label="First Name" />
```
scotttball: I think we have a firstName in the `en.js`

`<TextInput dataKey="person.firstName" label={__("first name")} />`

Same for line 21.

scotttball:
I'm trying to find out why the build is failing. It's due to a error after running `yarn check-format`. I'm looking into it right now.

Good code though. Take a look at my comments and give me your thoughts and then I think it's good.

joeldenning:
> I'm trying to find out why the build is failing. It's due to a error after running yarn check-format. I'm looking into it right now.

`yarn check-format` verifies that you didn't skip the pre-commit hook that runs prettier. The build says that `frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.web.component.js` does not match the prettier format. This could result from either (1) doing `git commit --no-verify` or (2) by your version of git not working well with husky. I've seen (2) happen before inside of docker containers or on Windows machines. To fix, run the following comand: `npx prettier 'src/**/*' --write`

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.pdf.component.js
```diff
@@ -1,5 +1,127 @@
...
+        />
+        <PositionedString dataKey="person.email" left="11.65%" top="28.1%" />
+        <PositionedCheckmark
+          dataKey="person.petitionerRepresentative"
+          left="20%"
+          top="31.5%"
+          shouldShow={data.person.petitionerRepresentative === "petitioner"}
+        />
+        <PositionedCheckmark
+          dataKey="person.petitionerRepresentative"
```
joeldenning: If you provide the `shouldShow` prop, the `dataKey` prop is ignored. This line (and similar ones in the other places where you're using PositionedCheckmark) can be deleted:

```suggestion
```

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.pdf.component.js
joeldenning: I like how you wrote this function. I think it's pretty easy to read and follow. One question is whether a "Fourth Degree" ever exists for any felonies/misdemeanors? I have never heard of that, but if that exists then our logic will think the fourth degree is actually a third degree. Might be worth checking with @tuckersamuelsen on this one.

joeldenning:
Also created https://github.com/JustUtahCoders/utahexpungements.org/pull/123 which might fix some of the prettier stuff

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.pdf.component.js
benschinn: Yes, it would be nice to get @tuckersamuelsen input on this. I found this [doc](https://www.utcourts.gov/howto/criminallaw/penalties.asp) that outlines classifications of criminal offenses in Utah. It appears there are these degrees of felonies:
1. Capital
2. First
3. Second
4. Third

Question **(1)** in [motion to reduce conviction](https://www.utcourts.gov/howto/criminallaw/docs/01_Motion_to_Reduce_Conviction.pdf) gives 2 options; your either requesting reduce your conviction by **one** or **two** degrees, there is no option for reduce your conviction by **three** degrees. This made me think that perhaps this from only pertains to people with either first, second, or third degree felonies?

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.web.component.js
```diff
@@ -1,5 +1,83 @@
...
+        dataKey="case.reducedConvictionDegree"
+        label={__("requesting court to reduce my convition to")}
+        options={convictionDegreeOptions}
+      />
+    </FormThatPrints>
+  );
 }
+
+const convictionDegreeOptions = [
+  { label: "First degree", value: "First Degree" },
```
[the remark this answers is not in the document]
benschinn: Yes, I agree that it would simply the code if we used an int value rather a string. I have no problem with it as long as it doesn't effect process-ability of the form.  I imagine it wouldn't.

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.web.component.js
benschinn: great catch! thanks!

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.pdf.component.js
benschinn: Ah, great to know. Thanks. I deleted dataKey and added debugKey prop.

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.pdf.component.js
joeldenning: Yeah I am unaware of how Capital offenses are treated in the reduction of conviction process. It seems unlikely that it would be an easy process to reduce a capital offense.

How are capital offenses being treated right now by your code?

tuckersamuelsen: Sorry late to this.  Here's some general info about the 402 process.

402s also apply to misdemeanors and infractions, so there's really 8 classifications: Capital, Felony 1, 2, and 3, Misdemeanor A, B, C, and infraction, in that order.  We'll never have to worry about Capital so we could probably just exclude that.

You can reduce it by a *maximum* of two steps - i.e., Felony 1 to Felony 3, or Felony 3 to Misdemeanor B, or MB to Infraction, etc.  Even then, the prosecutor has to specifically sign off to reduce by two steps - it's tougher and rarer, but it does happen and it never hurts to try.

Misdemeanor reductions are very important for expungement eligibility.  You have a lifetime limit of convictions before you aren't eligible, but if you can reduce a few infractions to misdemeanors than you might be.

We'll probably want to ask the user if they request a hearing, but generally they would probably prefer not to.  In Section B, you'd actually want to check both boxes if they want a 2-step reduction - that tells the court they'd like a 2-step but will settle for a 1-step.

benschinn: Amazing. Thanks @tuckersamuelsen. so exclude capital from this form.

Currently I only have the 3 degrees of felonies as options. 

I will add the additional 4 misdemeanors to the options on the web form. 

I will also work on adding validation logic for this (I think if the petitioner has a 1st degree felony he should only have  2nd, and 3rd degree felonies as options, not the entire 7 classifications of conviction).

tuckersamuelsen: Yes, that's correct.  It's never possible to reduce by 3 steps but it might confuse the user to have additional options.  So it should only allow them to choose 1- and 2-steps.

Thanks!

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.web.component.js
```diff
@@ -1,5 +1,140 @@
...
+  ] = React.useState(defaultConvictionDegreeOptions);
+
+  React.useEffect(() => {
+    setCurrentConvictionDegree(data.case.convictionDegree);
+  }, [data.case.convictionDegree]);
+  React.useEffect(() => {
+    if (currentConvictionDegree) {
+      setReducedConvictionDegreeOptions(
+        reducedConvictionDegreeOptionsFromCurrentConviction(
+          currentConvictionDegree
```
benschinn: Here is how I am dynamically setting the conviction degree options (i.e. if petitioner has a _first degree_ felony, his/her options would only be _second_ and _third degree_):

1. petitioner's conviction is set in component state as `currentConvictionDegree`
2. `currentConvictionDegree` is passed to `reducedConvictionDegreeOptionsFromCurrentConviction` func in useEffect
3.  `reducedConvictionDegreeOptionsFromCurrentConviction` returns correct conviction degree options based on the petitioner's conviction degree (e.g. if petitioner conviction degree is `'First Degree'`, valid options would be `'Second Degree'` and `'Third Degree'`.

 I am terrible at naming things. 🤦🏻‍♂️ I know it's verbose, but wanted it to be descriptive. Let me know if there is a better name.

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.web.component.js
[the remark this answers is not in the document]
benschinn: @scotttball After discovering 4 additional conviction classifications I re-introduced `degreeStringToInt` func. Initially I was only handling 1st, 2nd, and 3rd degree felonies, and it made sense that `degreeStringToInt` was not needed in that case. Now, with the additional misdemeanor classes A, B, C, and Infraction, `degreeStringToInt` function helps calculate difference of these additional convictions to another conviction.

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.web.component.js
joeldenning: I think it would be better not to hold this in state, but just to calculate it during render. It would simplify things a fair amount, including (if I'm reading this correctly) getting rid of both state values and both effects.

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.pdf.component.js
```diff
@@ -1,5 +1,136 @@
...
+        )}
+      </RenderPage>
+    </>
+  );
+}
+
+function requestedConvictionDegreeInformation(degA, degB) {
+  const degree = degreeStringToInt(degA);
+  const requestedDegree = degreeStringToInt(degB);
+  const isReduceByOneDeg = requestedDegree - degree > 1 ? false : true;
```
joeldenning: We could just use a conditional operator instead of ternary to get this:

```suggestion
  const isReduceByOneDeg = requestedDegree === 1
```

frontend/app/forms/motion-to-reduce-conviction/motion-to-reduce-conviction.web.component.js
benschinn: Thanks. I was torn on whether to get component state involved here or not. I will refactor this later tonight.

PR #126, opened 2019-11-12
fixed data keys on prior sections & completed section 4. 

oliviaclyde:
Fixed data keys on prior sections & completed section 4.
Proof of Completed Service - web components form complete.

### DIFF
--- a/frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
+++ b/frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
@@ -113,40 +113,40 @@ export default function ProofOfCompletedService_Web({ data }) {
           listed in section 3 by mail or commerical courier service to:
         </p>
         <TextInput
-          dataKey="proofofcompletedservice_person.fullname"
+          dataKey="person.proofOfCompletedServiceMailFullname"
           label="Name of Addressee"
         />
         <TextInput
-          dataKey="proofofcompletedservice_person.addressStreet"
+          dataKey="person.proofOfCompletedServiceMailAddressStreet"
           label="Street Address"
         />
         <TextInput
-          dataKey="proofofcompletedservice_person.addressCity"
+          dataKey="person.proofOfCompletedServiceMailAddressCity"
           label="City"
         />
         <TextInput
-          dataKey="proofofcompletedservice_person.addressState"
+          dataKey="person.proofOfCompletedServiceMailAddressState"
           label="State"
         />
         <TextInput
-          dataKey="proofofcompletedservice_person.addressZip"
+          dataKey="person.proofOfCompletedServiceMailAddressZip"
           label="Zip Code"
         />
         <p className="web-form-input">
           I have attached a signed receipt proving delivery. It was signed by:
         </p>
         <Checkbox
-          dataKey="proofOfCompletedService_case.addressee"
+          dataKey="case.proofOfCompletedServiceAddressee"
           label="the addressee personally"
         />
         <Checkbox
-          dataKey="proofOfCompletedService_case.someoneElse"
+          dataKey="case.proofOfCompletedServiceAuthorizedPerson"
           label="someone authorized by appointment or by law to receive service of process on behalf of the addressee"
         />
       </Section>
       <Section name="3. Service by Third Person">
         <Checkbox
-          dataKey="proofOfCompletedService_case.other"
+          dataKey="case.proofOfCompletedServiceThirdPerson"
           label="I am over the age of 18, and"
         />
         <p className="web-form-input">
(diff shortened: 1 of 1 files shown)


frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
```diff
@@ -113,40 +113,40 @@ export default function ProofOfCompletedService_Web({ data }) {
           listed in section 3 by mail or commerical courier service to:
         </p>
         <TextInput
-          dataKey="proofofcompletedservice_person.fullname"
+          dataKey="person.proofOfCompletedServiceMailFullname"
```
HappyViki: Why is that (and the other addressee fields) `person. ...` and not `case. ...`? I'm still learning what the conventions are, and wanted to know why you chose it.

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
```diff
@@ -113,40 +113,40 @@ export default function ProofOfCompletedService_Web({ data }) {
...
           label="Name of Addressee"
         />
         <TextInput
-          dataKey="proofofcompletedservice_person.addressStreet"
+          dataKey="person.proofOfCompletedServiceMailAddressStreet"
```
lukeschunk: same here i believe you can use person.addressStreet etc. Unless I'm wrong and there's a reason this should be a different data item

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
```diff
@@ -161,7 +161,107 @@ export default function ProofOfCompletedService_Web({ data }) {
...
+        <br></br>
+        <strong>OR</strong>
+        <Checkbox
+          dataKey="person.authorizedAgent"
+          label="an agent authorized by appointment or by law to receive service of
+          process and by also mailing a copy of the complaint and summons to the
+          named party, if the agent is one authorized by statute to receive process
+          and the statute so requires."
+        />
+        <br></br>
```
lukeschunk: This is a bit of a nitpick, but I prefer to not use too many `<br />` tags; if you can use styling here I think that would be better. you could use classnames and css like this: https://github.com/JustUtahCoders/utahexpungements.org/blob/master/frontend/app/forms/bank/fillable-form.component.js#L45

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
oliviaclyde: I think in this case, we need the person to be specific to this form. This part of the form is asking who the documents were mailed to. These fields should not be autopopulated with the same name as the plaintiff/petitioner. With this in mind, would the way I coded it be appropriate or would modification still be required?

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
oliviaclyde: As I indicated above, I believe this person to whom the court document was mailed is not the same person as the plaintiff/petitioner. So, I don't believe we want it to key off the same person? Maybe I'm thinking about this wrong and it won't auto-populate the same name. Is the ```person``` data key just to align formatting for all names, or is it tied to one name (like plaintiff) across all forms?

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
joeldenning: This information is specific to the case. See https://www.utcourts.gov/howto/service/docs/1020GE_Proof_of_Service.pdf where it asks for one case number for this information

image

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
joeldenning: > I believe this person to whom the court document was mailed is not the same person as the plaintiff/petitioner

Yes, this is correct ^ we should not be using `person.addressStreet`

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
```diff
@@ -113,40 +113,40 @@ export default function ProofOfCompletedService_Web({ data }) {
...
           label="Name of Addressee"
         />
         <TextInput
-          dataKey="proofofcompletedservice_person.addressStreet"
+          dataKey="person.proofOfCompletedServiceMailAddressStreet"
           label="Street Address"
         />
         <TextInput
-          dataKey="proofofcompletedservice_person.addressCity"
+          dataKey="person.proofOfCompletedServiceMailAddressCity"
```
joeldenning: Same here about this being specific to the case, not the person

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
joeldenning: Agreed on this. CSS is preferred over `<br>` when possible

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
```diff
@@ -161,7 +161,107 @@ export default function ProofOfCompletedService_Web({ data }) {
...
+        <br></br>
+        <small>If serving a city or town</small>
+        <Checkbox dataKey="person.recorder" label="the city/town recorder" />
+        <br></br>
+        <small>If serving a county</small>
+        <Checkbox dataKey="person.clerk" label="the county clerk" />
+        <br></br>
+        <small>If serving the state</small>
+        <Checkbox
+          dataKey="person.attorneyGeneral"
```
joeldenning: If I'm reading the form right, this is not actually the name of the attorney general. It's the name of a person who was served the document **in addition to the attorney general**. Additionally, this should be on `case`, not person, since the served documents are specific to this case.

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
```diff
@@ -161,7 +161,107 @@ export default function ProofOfCompletedService_Web({ data }) {
...
+          label="the attorney general, and to"
+        />
+        <TextInput
+          dataKey="person.proofOfCompletedServicePersonStatue"
```
joeldenning:
```suggestion
          dataKey="person.proofOfCompletedServicePersonStatute"
```

Please fix the other typos where it says Statue instead of Statute

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
oliviaclyde: I believe lines 178-184 is the individual the documents were served to followed by the designation "who is..." And this box would indicate that person is the attorney general. The text field after this in line 237 would be an add'l person required by the statute.

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
oliviaclyde: Good catch!

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
joeldenning: Ah you are completely right. I thought this was a TextInput, not a Checkbox. We definitely need the checkbox for this. Disregard what I said :)

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
oliviaclyde: Corrected the code to use `case` datakey.

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
oliviaclyde: Corrected to use `case` data key.

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
oliviaclyde: Added css styling.

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
oliviaclyde: Corrected to use `case` data key.

frontend/app/forms/proof-of-completed-service/proof-of-completed-service.web.component.js
oliviaclyde: Corrected any/all instance of statute.

Issue #101, opened 2019-10-29
Fix bug where sometimes PDF fails to process the first time


joeldenning:
For some of the example pdfs, they occasionally fail to process the first time you click on the button and you have to press the button again to process them.

As far as I can tell, this is related to something weird related to pdf parse

https://github.com/JustUtahCoders/utahexpungements.org/blob/master/backend/apis/parse-docket-pdf.api.js#L31

I think the easiest solution might be just to retry the parse inside of the api if the first one fails. And see if that provides good enough reliability of extracting the pdf text to live with.

PR #127, opened 2019-11-12
Add wait time for Busboy in parse-docket-pdf api (#101)

HappyViki:
Added wait time (highWaterMark: 2MiB buffer) for large files to process through Busboy (parse incoming HTML form data)
Fixed example problem: file with 33KB would sometimes not parse, and now it 99% does parse.

### DIFF
--- a/backend/apis/parse-docket-pdf.api.js
+++ b/backend/apis/parse-docket-pdf.api.js
@@ -7,6 +7,7 @@ app.post("/api/docket-pdfs", (req, res) => {
   let busboy;
   try {
     busboy = new Busboy({ headers: req.headers });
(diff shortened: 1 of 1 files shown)


backend/apis/parse-docket-pdf.api.js
```diff
@@ -59,3 +60,15 @@ app.post("/api/docket-pdfs", (req, res) => {
 
   return req.pipe(busboy);
 });
+
+function retryBusboy(data, remainingAttempts) {
+  //console.log('retrying', remainingAttempts)
```
jamesschlader: Go ahead and delete this console.log.

PR #129, opened 2019-11-13
Parse proceedings Resolves #97

### LINKED ISSUE #97: Calculate the "relevant date" for expungement eligibility.
See Tucker's comments on https://drive.google.com/drive/u/0/folders/1cW6RM4jLDeiLeSy7emI8_xdf6HhpfBhG

jamesschlader:
First draft at parsing the proceedings section. Adjusted the tests to standard format.


Cuthbert20:
Damn James These test look Nice! I did noticed what we talked about Wednesday Night about the description block causing an issue.

jamesschlader:
@Cuthbert20 What do you mean about the description block causing an issue? Could you expand on that a little?

Cuthbert20:
I was referring to the description segment of the tests.      

expect(charge2.description).toBe(
        "SPEEDING 70 in a 65 Class C Misdemeanor Offense Date: Month 00, 0000"
      );
Will break now that description is sliced into two sections offenseName & severity in the parsing phase.

oliviaclyde:
In our discussion today we discovered when copying and pasting pdf dockets you must first do: `exports.test = ` and put it inside back ticks. Copy from the first character in the pdf to the bottom. This will preserve the original case formatting.

Cuthbert20:
Hey what if we changed the file names from
test-pdf-1.js
test-pdf-1.test.js

to the first initial of the last name along with the last 4 of the case number. Since we are creating test cases for each pdf file, might make it easier to find out what pdf files have already been pulled in and have a test associated with the file. 
exp
test-pdf-e2266.js
test-pdf-e2266.test.js

@joeldenning would you be ok with this or do you think this is exposing to much personal information?

oliviaclyde:
Like @jamesschlader and I discussed, when returning keywords from the "proceedings" section, only the keyword and the date are returned (i.e. 6/30/16 Sentenced) There are other text strings before and after that keyword, but they appear to be clerk comments or notations. @joel or @tuckersamuelsen would it be important to return a certain number of characters before or after the keyword? It appears the clerk comments are more descriptive of actions taken or next steps in the process. Comments vary widely. Would returning this add'l information be important or not necessary?

Example: 

Defendant present with counsel
 Defendant admits to allegations knowingly and voluntarily made
 Sanction-Revoke PIA-Entered conviction
 **Sentenced**
 Apply $ in trust to fine & restitution and refund extra to payee
 Change Attorney fees to what had already been paid
 Defendant completed treatment

joeldenning:
This PR is not progressing, but we need it to. If we can't make progress on it in the next few days, I'll jump in to try to push it forward.

oliviaclyde:
As I've been reading court dockets, I think we're not accounting for probation. This needs to be included as a keyword we return. The waiting period could depend on when probation is complete.

(77-40-105 3(c)) "the following time periods have elapsed from the date the petitioner was convicted or released from incarceration, parole, or probation, whichever occurred last, for each conviction the petitioner seeks to expunge:...."

We are returning the "disposition date" which is the date the court made a final ruling in the case (i.e. guilty plea entered) but we should also be returning probation period or date probation was terminated successful. Whichever date occurred last is what the waiting period is based off. We'll need this information when we start writing the flags.

oliviaclyde:
For example: we need a way to parse out more than just "Probation" because it would return any instance of "Adult Probation and Parole". We really need to return the date when "probation as successful" is found. In the following example, however, I found this phrase runs onto 2 lines. We may need to return the line the keyword is found on + the following line.

01-01-01 Minute Entry - Minutes for APP REQUEST FOR TERMINATION
 The State moves to terminate the defendant's **probation
 as successful** as the defendant has completed all terms
 of probation. The Court grants the motion and executes
 the Order of Termination in open court.

jamesschlader:
@oliviaclyde good point about probation. I wonder whether the phrase "Order for Termination" might cover more ground if we search for it instead? Also, could you provide a link to the case you reference so that it could be added as a test case? I'll need a good case to test against in order to parse multi-line hits.

oliviaclyde:
https://drive.google.com/drive/folders/1tkwRAzGWXTb4rPwK6bkmUUbUnJ38lEjc

Issue #120, opened 2019-11-08
frontend/app/user/cases-list.component.js wire delete button


scotttball:
Ha good point, yeah we should fix that.

frontend/app/user/cases-list.component.js

* wire delete button up

_Originally posted by @joeldenning in https://github.com/JustUtahCoders/utahexpungements.org/pull/119_

scotttball:
Fixed by #147

PR #147, opened 2019-11-23
Wired delete button (#120)

HappyViki:
Location: /app/dashboard
Before: delete button didn't do anything.
After when button clicked:
- deletes case from database
- deletes case from browser for seamless UI
- if current case is deleted, `context.activePerson` and `context.activeCase` get updated to `""`
image

### DIFF
--- a/frontend/app/user/people-list.component.js
+++ b/frontend/app/user/people-list.component.js
@@ -1,6 +1,6 @@
 import React from "react";
 import { database } from "../../firebase";
-import { groupBy, partial, remove } from "lodash";
+import { groupBy, partial } from "lodash";
 import context from "../../context";
 import Accordion from "../utils/accordion.component.js";
 import CasesList from "./cases-list.component.js";
--- a/frontend/app/user/people-list.component.js
+++ b/frontend/app/user/people-list.component.js
@@ -1,6 +1,6 @@
 import React from "react";
 import { database } from "../../firebase";
-import { groupBy, partial } from "lodash";
+import { groupBy, partial, remove } from "lodash";
 import context from "../../context";
 import Accordion from "../utils/accordion.component.js";
 import CasesList from "./cases-list.component.js";
@@ -32,6 +32,20 @@ export default class PeopleList extends React.Component {
     });
   };
 
+  deleteCase = (person, kase, e) => {
+    e.preventDefault();
+    const array = [...this.state.cases];
+    remove(array, item => {
+      return item === kase;
+    });
+    this.setState({ cases: array });
+    context.setContext({
+      activePerson: "",
+      activeCase: ""
+    });
+    database.deleteCase(kase.id);
+  };
+
   render() {
     const { cases, people } = this.state;
     const { activeCase } = this.props;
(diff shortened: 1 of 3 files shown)


frontend/app/user/people-list.component.js
```diff
@@ -32,6 +32,20 @@ export default class PeopleList extends React.Component {
     });
   };
 
+  deleteCase = (person, kase, e) => {
```
joeldenning: This function could be simplified to the following:

```js
deleteCase = (person, kase, e) => {
  e.preventDefault()
  this.setState(prevState => ({
    cases: prevState.cases.filter(c => c !== kase)
  }, () => {
    database.deleteCase(kase.id)
  })
}
```

You don't need lodash's remove function, you can just use [Array.prototype.filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter). Additionally, in React class components you should use `prevState` with an updater function instead of `this.state` [as described here](https://reactjs.org/docs/react-component.html#setstate) when updating state based on it's previous value.

frontend/app/user/people-list.component.js
```diff
@@ -32,6 +32,20 @@ export default class PeopleList extends React.Component {
...
+    e.preventDefault();
+    const array = [...this.state.cases];
+    remove(array, item => {
+      return item === kase;
+    });
+    this.setState({ cases: array });
+    context.setContext({
+      activePerson: "",
```
joeldenning: Why do we want to reset the activePerson when a case is deleted?

frontend/app/user/people-list.component.js
```diff
@@ -32,6 +32,20 @@ export default class PeopleList extends React.Component {
...
+      activeCase: ""
```
joeldenning: Do we always need to reset the active case when a case is deleted? Can't we keep the activeCase if it wasn't the one that was deleted?

HappyViki: We want to reset `activePerson` and `activeCase` only when active case is deleted. I'll add in a statement that checks for that.

frontend/app/user/people-list.component.js
HappyViki: I suppose only `activeCase` can be reset, because in `frontend/app/user/dashboard.component.js`, only when both `activeCase` and `activePerson` are truthy it displays the info:
```
{context.activePerson && context.activeCase && (
  <p>
    You are currently working on case{" "}
    <strong>{context.activeCase.name}</strong> for{" "}
    <strong>{context.activePerson.name}</strong>.
  </p>
)}
```

frontend/app/user/people-list.component.js
joeldenning: > cases: prevState.cases.filter(c => c !== kase) makes it look like the state is being changed directly, but it isn't, because it's in setState method.

No. The reason has nothing to do with it being inside of a setState method. It has everything to do with Array.prototype.filter creating a **new array** without modifying the existing one. You can use filter safely because it doesn't mutate the array

frontend/app/user/people-list.component.js
joeldenning: 👍

PR #152, opened 2019-12-17
removed extra side scroll

MeghanBomberger:
We were getting an unneeded side to side scroll


lukeschunk:
where you seeing this? I just tried a couple different browser sizes and didn't see it (mac chrome)

MeghanBomberger:
image

I'm on Chrome, but with Windows rather than Mac.

PR #160, opened 2020-01-25
initial work on 1st degree felony flag

oliviaclyde:
I need some help on the logic for the first degree felony flag.

### DIFF
--- a/backend/apis/flag-utils/is-first-degree-felony-flag-util.js
+++ b/backend/apis/flag-utils/is-first-degree-felony-flag-util.js
@@ -1,19 +1,8 @@
 exports.isFirstDegreeFelony = function isFirstDegreeFelony(parsedObj) {
-  if (
-    parsedObj.charges
-      .map(charge => charge.severity.toLowerCase())
-      .includes("1st degree felony")
-  ) {
-    if (
-      parsedObj.charges
-        .map(charge => charge.severity.toLowerCase())
-        .includes("amended")
-    ) {
-      return false;
-    } else {
-      return true;
-    }
-  } else {
-    return false;
-  }
+  const felonies = parsedObj.charges.filter(
+    charge =>
+      charge.severity.toLowerCase().includes("1st degree felony") &&
+      !charge.severity.toLowerCase().includes("amended")
+  );
+  return !!felonies.length;
 };
--- a/backend/apis/flag-utils/is-first-degree-felony-flag-util.js
+++ b/backend/apis/flag-utils/is-first-degree-felony-flag-util.js
@@ -1,26 +1,19 @@
 exports.isFirstDegreeFelony = function isFirstDegreeFelony(parsedObj) {
-  // if (
-  //   parseObj.charges.map(charge => charge.severity.toLowerCase()).includes("1st degree felony" && "amended")
-  // ){
-  //   return true;
-  // } else if (
-  //   parseObj.charges.map(charge => charge.severity.toLowerCase()).includes("1st degree felony")
-  // ){
-  //   return severity;
-  // } else {
-  //   return false
-  // }
-
-  console.log(parsedObj);
-
   if (
-    parsedObj.charges[0].severity.toLowerCase().includes("1st degree felony") &&
-    parsedObj.charges[0].severity.toLowerCase().includes("(402 amended)")
+    parsedObj.charges
+      .map(charge => charge.severity.toLowerCase())
+      .includes("1st degree felony")
   ) {
-    return true;
+    if (
+      parsedObj.charges
+        .map(charge => charge.severity.toLowerCase())
+        .includes("amended")
+    ) {
+      return false;
+    } else {
+      return true;
+    }
   } else {
-    return parsedObj.charges[0].severity
-      .toLowerCase()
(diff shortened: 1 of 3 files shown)


backend/apis/flag-utils/is-first-degree-felony-flag-util.js
```diff
@@ -0,0 +1,12 @@
+exports.isFirstDegreeFelony = function isFirstDegreeFelony(parsedObj) {
+  if (
```
oliviaclyde: Do we need to iterate through "charges" if there is more than 1 charge on the docket? May need to add a for loop to do this?

benschinn: You can iterate through the `charges` array using `map` function. You can use the map method to return a new array that with only the value of the `severity` key and see if the the new array includes string "1st degree felony" or "amended".

Something like this:
```js
parseObj.charges.map(charge => charge.severity.toLowerCase()).includes("1st degree felony" || "amended")
```

benschinn: Contrived example but hopefully helps understand the usage of [`Array.map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) here:

```js
const people = [{name: 'charley', age: 12}, {name: 'xavier', age: 23}, {name: 'tina', age: 11}]

const onlyNames = people.map(person => person.name)
//["charley", "xander", "luna"]

onlyNames.includes("charley" || "tina")
//true
```

oliviaclyde: Thanks for the help! I think I have it now.

oliviaclyde:
I have updated the logic and used a nested "if" to check for "1st degree felony" and if true, then check for "amended". 

My concern is we also need to check the disposition field on each of these flags to make sure it doesn't return "dismissed"? I believe that was a function @benschinn was writing so we could call it on each of our flags. I believe if a change is dismissed, then it doesn't count against them, right @tuckersamuelsen?

oliviaclyde:
@tuckersamuelsen I can't find a docket that has a charge of "1st Degree Felony" that hasn't been amended down. I'd like to get one so I can make sure my flag returns "true" on that case.

backend/apis/flag-utils/is-first-degree-felony-flag-util.js
```diff
@@ -0,0 +1,19 @@
...
+        .map(charge => charge.severity.toLowerCase())
+        .includes("amended")
+    ) {
+      return false;
+    } else {
+      return true;
+    }
+  } else {
+    return false;
+  }
```
lukeschunk: 💡 I might have a suggestion for how we can make this a little more concise. overall though the idea looks good to me; stand by for a suggestion

oliviaclyde: I'm happy to refactor if there is a more concise alternative. I'm new to Node, so I'm learning as I go. All suggestions will help me improve. Do you have a recommendation?

lukeschunk: What about something like this: 
```
const felonies = parsedObj.charges.filter(charge => (charge.severity.toLowerCase().includes("1st degree felony") && !charge.severity.toLowerCase().includes("amended"))
```
This will give you an array of all the charges that have "1 st degree felony" but that aren't amended. Then for the final return of the function you can do something like this: 
```
return !!felonies.length
```
If the felonies array has any items, it will return true, otherwise, 0 is falsey and it will return false. 

It doesn't have to be exactly that but hopefully that gives you something to go off, and I think it eliminates the bug where you're looping through the charges independently so the "amended" might not apply to the 1st degree felony. 

The whole func: 
```
exports.isFirstDegreeFelony = function isFirstDegreeFelony(parsedObj) {
  const felonies = parsedObj.charges.filter(charge => (charge.severity.toLowerCase().includes("1st degree felony") && !charge.severity.toLowerCase().includes("amended"))
  return !!felonies.length
}
```

oliviaclyde: Yes, this makes more sense. If I loop independently, then a lesser severity charge (for example Misdemeanor) could be "amended" and the code would consider the "amended" and return false. I thought this may have been avoided with the nested if statements I was using, but it was still cumbersome code.

oliviaclyde:
@lukeschunk I tested this and it looked okay on several dockets I tried, but when I used the new docket Tucker uploaded that contained several 1st degree felony charges (I'll PM you the case name) it returned false when it should have returned true. I wonder if this was because of the unusal formatting on the pdf....it also had a weird format on the raw parcer too. Can you take a look?

benschinn:
@Mengqi89 @oliviaclyde and I looked into the issue mentioned above, and it sounds like the issue is with the parser. Tuning the parser should be addressed in a separate PR, the code here LGTM.

lukeschunk:
@benschinn what is the issue with the parser? we could open an issue

benschinn:
@lukeschunk
https://github.com/JustUtahCoders/utahexpungements.org/issues/174

Issue #141, opened 2019-11-20
Misdemeanor B Flag


tuckersamuelsen:
Determine if the individual has at least 4 convictions where 3 are Misdemeanor B's.

joevus:
I can't assign myself to an issue yet, but I'd be happy to work on this one!

PR #161, opened 2020-01-27
add class b misdemeanor flag

joevus:
Addresses https://github.com/JustUtahCoders/utahexpungements.org/issues/141

I just put this up real quick, still need to review it, but thought I'd put it up here to get some eyes on it to make sure I'm going the right direction.

@benschinn are there any tests I should run to make sure nothing broke?
@tuckersamuelsen does a "Class B Misdemeanor" always show up with this capitalization?

Also, I'm first counting the _charges_ to check that there are 4 or more. Then I loop through the _charges_ to see if there are at least 3 that are Class B. The description says to check the _convictions_ rather than _charges_ so wanted to check if I'm doing that right.

### DIFF
--- a/backend/apis/flag-utils/is-class-b-misdemeanor.js
+++ b/backend/apis/flag-utils/is-class-b-misdemeanor.js
@@ -1,15 +1,12 @@
 exports.isClassBMisdemeanor = function(parsedObj) {
-  // Determine if the individual has at least 4 convictions where 3 are Misdemeanor B's
-  // must have at least 4 convictions
-  if (parsedObj.charges.length < 4) {
-    return false;
-  }
-  let numOfClassBMisdemeanors;
-  // 3 or more of the convictions must be Class B Misdemeanors
-  for (let i = 0; i++; i < parsedObj.charges.length) {
-    if (parsedObj.charges[i].includes("B Misdemeanor")) {
-      numOfClassBMisdemeanors++;
+  // check if at least one charge is a Class B Misdemeanor
+  let containsClassBMisdemeanor = false;
+
+  parsedObj.charges.forEach(x => {
+    if (x.severity.includes("B Misdemeanor")) {
+      containsClassBMisdemeanor = true;
     }
-  }
-  return numOfClassBMisdemeanors >= 3;
+  });
+
+  return containsClassBMisdemeanor;
 };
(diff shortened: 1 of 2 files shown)


backend/apis/flag-utils/is-class-b-misdemeanor.js
```diff
@@ -0,0 +1,15 @@
...
+  // must have at least 4 convictions
+  if (parsedObj.charges.length < 4) {
+    return false;
+  }
+  let numOfClassBMisdemeanors;
+  // 3 or more of the convictions must be Class B Misdemeanors
+  for (let i = 0; i++; i < parsedObj.charges.length) {
+    if (parsedObj.charges[i].includes("B Misdemeanor")) {
+      numOfClassBMisdemeanors++;
+    }
```
lukeschunk: 💡 i think you could use `.reduce` here and it would be a little cleaner/more concise.

benschinn:
As far as tests, the CI runs tests, so if any existing tests fail you should know. You bring up a great point though, we might want unit tests for our flag utils.

tuckersamuelsen:
> Addresses #141
> 
> I just put this up real quick, still need to review it, but thought I'd put it up here to get some eyes on it to make sure I'm going the right direction.
> 
> @benschinn are there any tests I should run to make sure nothing broke?
> @tuckersamuelsen does a "Class B Misdemeanor" always show up with this capitalization?
> 
> Also, I'm first counting the _charges_ to check that there are 4 or more. Then I loop through the _charges_ to see if there are at least 3 that are Class B. The description says to check the _convictions_ rather than _charges_ so wanted to check if I'm doing that right.

@joevus to my knowledge, yes.
Are you looking within 1 docket?  If yes, you'll only have to confirm the presence of a Misdemeanor B, and not add up charges.  It doesn't count multiple charges as multiple convictions, if that makes sense.

joevus:
@tuckersamuelsen Yes, I believe this is checking only one docket. So now I've changed the code to check for at least one charge with a class b misdemeanor on 1 docket. I guess the checking if there are 3 convictions will come later after they upload multiple dockets then.

@lukeschunk the code changed since I don't have to count class b misdemeanor charges anymore per @tuckersamuelsen's comment. Using a forEach method, but happy to take any suggestions to clean it up.

PR #166, opened 2020-02-13
Checking if automobile-homicide

Cuthbert20:
Created file to check if automobile-homicide && imported into generate-flag-json.utils.js

### DIFF
--- a/backend/apis/flag-utils/is-automobile-homicide.js
+++ b/backend/apis/flag-utils/is-automobile-homicide.js
@@ -1,0 +1,13 @@
+exports.isAutomobileHomicide = function(parsedObj) {
+  //Check if offense is a Automobile Homicide
+  let automobileHomicide = false;
+  parsedObj.charges.forEach(val => {
+    if (
+      val.offenseName.includes("AUTOMOBILE HOMICIDE") &&
+      val.offenseName.includes("DUI")
+    ) {
+      automobileHomicide = true;
+    }
+  });
+  return automobileHomicide;
+};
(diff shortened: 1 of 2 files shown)


backend/apis/flag-utils/is-automobile-homicide.js
```diff
@@ -0,0 +1,13 @@
...
+  let automobileHomicide = false;
+  parsedObj.charges.forEach(val => {
+    if (
+      val.offenseName.includes("AUTOMOBILE HOMICIDE") &&
+      val.offenseName.includes("DUI")
+    ) {
+      automobileHomicide = true;
+    }
+  });
+  return automobileHomicide;
```
lukeschunk: i think this looks good; the only thing I worry about here is that you're checking for an all caps "AUTOMOBILE HOMICIDE" this will not work if the docket happens to say something like `automobile-homicide` `auto-homicide` etc. 

@tuckersamuelsen might be able to speak to whether or not that will ever happen, or if it is always an all caps "AUTOMOBILE HOMICIDE" or "DUI" spelled out.

Cuthbert20: I can add additional conditionals to check for lowerCase I will also ask Tucker if the offense is always capped. I think he told me before it is but, I will double check.

Thanks!

lukeschunk: you shouldn't need to add conditionals, just lower case the val 👍

Cuthbert20: Sweet thanks

backend/apis/flag-utils/is-automobile-homicide.js
```diff
@@ -0,0 +1,13 @@
+exports.isAutomobileHomicide = function(parsedObj) {
+  //Check if offense is a Automobile Homicide
+  let automobileHomicide = false;
+  parsedObj.charges.forEach(val => {
+    if (
+      val.offenseName.toLowerCase().includes("automobile homicide") &&
+      val.offenseName.toLowerCase().includes("dui")
```
Cuthbert20: @lukeschunk converted string to lower case before using includes method.

PR #169, opened 2020-02-19
created isClassAMisdemeanor flag function

HappyViki:
I tried my best to create a function based off of the law. I'm pretty sure that we either need to be able to process multiple dockets at once or have a different counting system in place, other than true/false object.

**Law:**
> (b) any combination of three or more convictions
> other than for drug possession offenses
> that include two class A misdemeanor convictions,
> each of which is contained in a separate criminal episode;

**This is pseudo code I wrote that I based my final function on:**
```
Flag = FALSE

Convictions = {
class_a_misdemeanor:0,
plain:0
}

Convictions per docket:
Skip if "drug possession offense" found
Return plain++ if felony or any other conviction found
Return class_a_misdemeanor++ if "class A misdemeanor conviction" found

Multiple dockets:
Return Flag(TRUE) if TOTAL Convictions(3+) AND Convictions.class_a_misdemeanor(2+)
```

### DIFF
--- a/backend/apis/flag-utils/is-class-a-misdemeanor.js
+++ b/backend/apis/flag-utils/is-class-a-misdemeanor.js
@@ -1,0 +1,52 @@
+// "(b) any combination of three or more convictions
+// other than for drug possession offenses
+// that include two class A misdemeanor convictions,
+// each of which is contained in a separate criminal episode;"
+
+// Flag = FALSE
+
+// Convictions = {
+// class_a_misdemeanor:0,
+// plain:0
+// }
+
+// Convictions per docket:
+// Skip if "drug possession offense" found
+// Return plain++ if felony or any other conviction found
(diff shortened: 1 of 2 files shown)


backend/apis/flag-utils/is-class-a-misdemeanor.js
```diff
@@ -0,0 +1,52 @@
...
+// plain:0
+// }
+
+// Convictions per docket:
+// Skip if "drug possession offense" found
+// Return plain++ if felony or any other conviction found
+// Return class_a_misdemeanor++ if "class A misdemeanor conviction" found
+
+// Multiple dockets:
+// Return Flag(TRUE) if TOTAL Convictions(3+) AND Convictions.class_a_misdemeanor(2+)
```
lukeschunk: I don't know if all these comments are necessary in the long term but I'm ok to merge it especially while we're very much in a devlopment phase

PR #170, opened 2020-02-23
added child abuse flag, added engines field to package.json

Mengqi89:
I had to add "engines" field in package.json due to the following error on line 90 in parse-docket.pdf.utils.js

> Rest/spread properties are not supported until Node.js 8.3.0.


Mengqi89:
I am going to create a new pull request to see if I can avoid the travis-ci fail.

PR #173, opened 2020-03-02
support for handling multiple file uploads

benschinn:
It's not pretty, but it's working. Feedback welcome. We may also want to set a limit on number of files.

currently payload looks like:
```
[{"caseNumber": "123" ...}, {"caseNumber": "124" ...}]
```
should we change the payload to have some indication of what parsed data belongs to which docket? something like this? 
```
[{"fileName": "docket.pdf", "caseNumber": "123" ...}, {"fileName": "docket_1.pdf", "caseNumber": "124" ...}]
```

### DIFF
--- a/backend/apis/parse-docket-pdf.api.js
+++ b/backend/apis/parse-docket-pdf.api.js
@@ -6,7 +6,12 @@ const { generateFlagJson } = require("./generate-flag-json.utils");
 
 app.post("/api/docket-pdfs", (req, res) => {
   let busboy;
-  const busBoyConfig = { headers: req.headers };
+  const busBoyConfig = {
+    headers: req.headers,
+    limits: {
+      fileSize: 10
+    }
+  };
 
   try {
     busboy = new Busboy(busBoyConfig);
--- a/backend/apis/parse-docket-pdf.api.js
+++ b/backend/apis/parse-docket-pdf.api.js
@@ -6,8 +6,10 @@ const { generateFlagJson } = require("./generate-flag-json.utils");
 
 app.post("/api/docket-pdfs", (req, res) => {
   let busboy;
+  const busBoyConfig = { headers: req.headers };
+
   try {
-    busboy = new Busboy({ headers: req.headers });
+    busboy = new Busboy(busBoyConfig);
     busboy.highWaterMark = 2 * 1024 * 1024; // Set 2MiB buffer
   } catch (err) {
     console.error(err);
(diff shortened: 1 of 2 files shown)


backend/apis/parse-docket-pdf.api.js
```diff
@@ -6,8 +6,10 @@ const { generateFlagJson } = require("./generate-flag-json.utils");
 
 app.post("/api/docket-pdfs", (req, res) => {
   let busboy;
+  const busBoyConfig = { headers: req.headers };
```
benschinn: Here we can add additional configuration such as limiting number of files. Do we know the max number of dockets a user can upload?

backend/apis/parse-docket-pdf.api.js
```diff
@@ -25,6 +29,7 @@ app.post("/api/docket-pdfs", (req, res) => {
         .send({ error: "Uploaded file did not have mimetype application/pdf" });
     }
     fileWasUploaded = true;
+    if (filename) count++;
```
benschinn: busboy processes one file at a time. this is how I track the number of files were uploaded. I think messing with counters can be error prone. if anyone has a better solution, please share!

backend/apis/parse-docket-pdf.api.js
```diff
@@ -34,9 +39,13 @@ app.post("/api/docket-pdfs", (req, res) => {
             if (processJson) {
               //comment this back in if you want to test flag json
               //res.send(generateFlagJson(parsePdfText(thePdf.text)));
-              res.send(parsePdfText(thePdf.text));
+              payload.push(parsePdfText(thePdf.text));
+              count--;
+              if (count === 0) res.send(payload);
```
benschinn: also, do we prefer using if statement here or something like this instead `count === 0 && res.send(payload)` instead?

lukeschunk: i think this is fine 👍

backend/apis/parse-docket-pdf.api.js
benschinn: I think it's hard to set a limit on number of dockets. for now it is more reasonable to set limit on `fileSize`. It has been set to 10mb which should be plenty.

backend/apis/parse-docket-pdf.api.js
benschinn: no feedback. will resolve for now.

benschinn:
@lukeschunk could you merge it in? I don't have write access to the repo.

PR #182, opened 2020-04-22
Justice court cover sheet

Mengqi89:


Mengqi89:
Because now we have added the Justice Court Cover Sheet, it is probably better to refactor the District Court Cover Sheet which was named as "coversheet" throughout the code. What do you think?

Issue #183, opened 2020-04-29
Form Bank Issues


benschinn:
_"Hey all - just had a call with Tucker and the expungement navigator, Jake. It was super cool. Jake has actually been using the form bank portion of the site for several weeks.
In doing that, he’s come up with a handful of issues to address. If we can address them, it’s possible that Tucker’s company will link (or embed, etc) to our site. Which would be awesome for driving some traffic and potentially making a bigger impact."_ - Luke

To help with addressing the issues below, please follow these steps:
1. If you are actively taking on the issue please check the box.
2. Once your PR to address the issue has been merged in please do ~~strikethrough~~ on the issue and reference your MR link that addresses it. 

#### Issues:

2. - [x] ~~Can we remove the dates from the signature line being populated, so that they have to sign the date as well as their signature and location.~~ https://github.com/JustUtahCoders/utahexpungements.org/pull/186 https://github.com/JustUtahCoders/utahexpungements.org/pull/198
   1. - [x] ~~On the court address population for West Valley, its coming up as 3590 South 2700 West West Valley, UT. Let’s get a comma in between the city and the street addresses.~~ https://github.com/JustUtahCoders/utahexpungements.org/pull/185

3. - [x] ~~When I move to do the coversheet, it is not allowing me to input the defendant/respondent.~~ https://github.com/JustUtahCoders/utahexpungements.org/pull/190

4. - [x] ~~Also, there is a separate cover sheet for justice court cases, this is only allowing district court case.~~ https://github.com/JustUtahCoders/utahexpungements.org/pull/182

5. - [x] ~~Finally, on the district court case, the box “This case is exempt for URCP 26” needs to be checked o page one.~~ https://github.com/JustUtahCoders/utahexpungements.org/pull/193

6. - [x] ~~Phone number not populating from each form. Have to enter phone number each time. Ok wait, seems to only be the Order form and the Cover Sheet form.~~ https://github.com/JustUtahCoders/utahexpungements.org/pull/194

7. - [x] ~~REMOVE the check box on the Expungement Orders for whether the matter is being resolved by pleading or by hearing. The Judge will check that box. It is too early to tell. A hearing will be held if a victim or prosecutor objects. We don’t know if that is going to happen when we submit this form.~~  https://github.com/JustUtahCoders/utahexpungements.org/pull/197

[...]

PR #192, opened 2020-05-10
fix urls of page 4/5 in Proof of Completed Service

Mengqi89:
Fix wrong urls


Mengqi89:
> @Mengqi89 this looks great. does this MR address any of the items in #183 ?

Not really. I just happened to notice it.

PR #193, opened 2020-05-11
add checkmark on URCP26 exemption in district court cover sheet

Mengqi89:
This is issue 5 in https://github.com/JustUtahCoders/utahexpungements.org/issues/183


oliviaclyde:
Looks like this has been reviewed and approved. I'll merge it in.

Issue #202, opened 2020-06-11
Form Bank Issues rev2


oliviaclyde:
_Here is the rev2 version of form issues from Jake._

To help with addressing the issues below, please follow these steps:

If you are actively taking on the issue please check the box.
Once your PR to address the issue has been merged in please do ~~strikethrough~~ on the issue and reference your MR link that addresses it.

**Issues:**

1. [ ] Phone number still doesn’t want to populate between forms. This time between Conviction Petition and Motion to Waive Fees.
2. [ ] Tab button sometimes refreshes page and sets you back on the original case you ever entered.
3. [ ] Motion to Waive Fees needs an area to enter the Respondent on page 1
4. [ ] Order on Motion to Waive Fees: Respondent NOT defendant. Petitioners name still not populating.
5. [ ] Still need a Justice Court coversheet.
6. [ ] Phone number not populating on Acceptance of Service
7. [ ] Consent and Waiver of Hearing is same 2 pages.
8. [ ] When I click justice court for the first time, it always crashes the system.
9. [ ] Uncheck the box on expungement order “the pleadings and other parties”
10. [ ] On Order on Motion to Waive fees, get rid of auto populating on State of Utah. In fact, just get rid of populating the respondent altogether, no need
11. [ ] South Jordan Justice Court address is incorrect.
12. [ ] Page 2 of District Court Coversheet is Missing
13. [ ] Some forms are adding a comma after the city on the address line, others are not
14. [ ] Farmington District Court is missing
15. [ ] Order to waive Fees not populating phone number
16. [ ] Motion to Waive Fees does not populate employer’s address and phone number, job title, or hourly rate or annual salary
17. [ ] Does not provide subtotal for total monthly gross income
18. [ ] Does not provide subtotal for total monthly tax deductions, nor does it calculate after-tax monthly income
19. [ ] Does not provide subtotal for monthly expenses
20. [ ] Motion to waive populates across Cases, BRILLIANT BEAUTIFUL GOOD JOB! But the I am employed part does not populate across
21. [ ] American Fork District Court is populating the address as: Church St. and E Main St.(Approximate Address) 75 East 80 North, Suite 202American Fork, UT 84003-0986. Get rid of everything before 75 East 80 North, Suite 202American Fork, UT 84003-0986. It’s running onto a second line which is actually pushing down some of the pages somehow, so if needed abbreviate to 75 E. 80 N. Ste. 202American Fork, UT 84003-0986
[...]

PR #203, opened 2020-06-23
Fix phone number

Mengqi89:
Fix phone number population issue in Conviction Petition, Motion to Waive Fees, Acceptance of Service and all Orders.


Mengqi89:
This MR fixes 1, 6, 15, 35 and 43 in https://github.com/JustUtahCoders/utahexpungements.org/issues/202

PR #211, opened 2020-07-06
Remove judge name

Mengqi89:
Removed judge names next to signature lines (Issue 25 and 26 in https://github.com/JustUtahCoders/utahexpungements.org/issues/202)


benschinn:
We may want to remove the input for the judge name in the web component as well. Having an input that doesn't print the information may cause confusion.

Mengqi89:
> We may want to remove the input for the judge name in the web component as well. Having an input that doesn't print the information may cause confusion.
Just did.

Issue #174, opened 2020-03-04
Parser logic failing with difference in docket formatting


oliviaclyde:
Parser logic is failing due to docket formatting differences. It appears most dockets have indentations with white space inside each charge. Outlier case has a docket with left alignment on the charges section. This difference in alignment throws the parsing off and the "Offense Date" is parsed in the "Severity" section. The flag checking the severity of the charge then fails because the array returns empty of key words. 

Reference docket:  Axxxx  xxxxx6862

Mengqi89:
This is what the parser yields:
> {
  "caseNumber": "191906862",
  "charges": [
    {
      "statute": "76-5-202",
      "offenseName": "AGGRAVATED MURDER",
      "severity": "Offense Date: June 17, 2019"
    },
    {
      "statute": "76-5-302",
      "offenseName": "AGGRAVATED KIDNAPPING",
      "severity": "Offense Date: June 17, 2019"
    },
    {
      "statute": "76-8-306(1)",
      "offenseName": "OBSTRUCTING JUSTICE",
      "severity": "Offense Date: June 17, 2019"
    },
    {
      "statute": "76-9-704",
      "offenseName": "ABUSE OR DESECRATION OF A DEAD HUMAN BODY",
      "severity": "Felony"
    }
  ],
  "accountSummary": [
    {
      "name": "ACCOUNT SUMMARY",
      "collection": false
    }
  ]
}

PR #220, opened 2020-08-22
Restructured file with a lot of regex

HappyViki:
Issue: https://github.com/JustUtahCoders/utahexpungements.org/issues/174

Problem: Parser doesn't always parse what's needed.

Solution: Completely restructure the parsing file to use mostly just Regex.

Before: 
1. File would parse line by line
2. Use minimal regex to determine how to display info in line

After: 
1. File parses in chunks based on regex
2. Iterates over multiple similar chunks
3. Uses a lot of regex to refine search on those chunks

Note: I wanted to add some extra stuff, but I feel like it's best to push something that works, rather than try and perfect it. I modified the test a bit. My code passes the test now. I would still like someone to review with me and make sure I've covered all the defaults. I'm afraid that not everything has been covered in the test, and so I may have missed something.

### DIFF
--- a/backend/apis/normal-docket-pdf.test.js
+++ b/backend/apis/normal-docket-pdf.test.js
@@ -23,22 +23,20 @@ describe("normal docket pdf file", () => {
     let actual1 = parsedText.charges[1];
 
     expect(actual0.statute).toBe("53-3-202(1)(A)");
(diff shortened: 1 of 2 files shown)


benschinn (approved):
@HappyViki this looks amazing! thanks for working on this.

